PR #414, opened 2026-05-08
Fix/suspension/can node launch fix

BohdanZasymovych:
## Description
The launch file for the can sender and receiver nodes was updated to follow the best practices. Check of whether the can interface is up was added; nodes lifecycle handling updated to avoid race conditions. The command which starts up the CAN interface added to the `entrypoint.bash`.

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- The launch file `can.launch.py` fully refactored to handle errors and avoid race conditions in the nodes lifecycle handling.
- The `Dockerfile` and `entrypoint.bash` updated to start up the CAN interface (set name and bitrate); repeated installation of the package in the `Dockerfile` removed.
- Added `README.md` describing the current setup of the can, environmental variables, launch parameters.
- The `docker-compose-example.yaml` updated.

## Testing
- Container successfully builds
- On the run of the container can starts up correctly
- The package successfully builds
- The nodes successfully starts

Because i had only one USB to CAN converter i wasn't able to verify if receiving and transmitting works correctly. Full tests need to be conducted on the motors or with two converters. Until the tests are conducted the pull request will not be merged.

## Additional Notes

Setting up the CAN interface was not done in the launch file since the command for it requires sudo privileges and in my opinion launching the launch file with sudo will not be convenient. Instead i added it to the `entrypoint.sh` and now it starts up when the container stats up. In the future on the Jetson start up of the interface can be made automatic on the boot of the Jetson itself fully bypassing docker, ros and does not requiring any command to run manually.


docker/Dockerfile
```diff
@@ -18,10 +18,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     ros-${ROS_DISTRO}-teleop-twist-keyboard \
     usbutils \
     iproute2 \
+    sudo \
     pkg-config \
     g++ make cmake \
     && rm -rf /var/lib/apt/lists/*
 
+    RUN echo "ALL ALL=(ALL) NOPASSWD: /usr/sbin/ip" >> /etc/sudoers
```
OShevchuk-UCU: Remove the redundant leading spaces before `RUN`; Dockerfile instructions should start at column 0.

docker/entrypoint.bash
```diff
@@ -2,11 +2,36 @@
...
+
+# -------------------- CAN Interface Setup --------------------
+
+echo "[CAN] INFO: Checking for CAN hardware on interface: ${TARGET_IFACE}"
+
+if ip link show "${TARGET_IFACE}" > /dev/null 2>&1; then
+    echo "[CAN] INFO: Hardware found. Configuring ${TARGET_IFACE} at ${TARGET_BITRATE} bps..."
+
+    if ip link show "${TARGET_IFACE}" | grep -q "UP"; then
+        sudo ip link set "${TARGET_IFACE}" down || true
```
OShevchuk-UCU: Use the same absolute ip path in both sudoers and entrypoint, or remove sudo entirely if the container already runs as root.

src/indomitus_rover_bringup/launch/can.launch.py
OShevchuk-UCU, line 1: Lifecycle event handlers were registered after the lifecycle nodes were added to the launch graph, so fast-starting processes could miss OnProcessStart transitions and fail to configure deterministically. The handlers are now registered before node startup, eliminating the lifecycle startup race.

```text
+++ b/src/indomitus_rover_bringup/launch/can.launch.py
@@ -156,6 +156,13 @@ def generate_launch_description() -> LaunchDescription:
         output="screen",
     )
 
+    sender_configure_handler = _auto_configure(sender_node, "socket_can_sender")
+    receiver_configure_handler = _auto_configure(receiver_node, "socket_can_receiver")
+    sender_activate_handler = _auto_activate(sender_node, "socket_can_sender")
+    receiver_activate_handler = _auto_activate(receiver_node, "socket_can_receiver")
+    sender_fail_fast_handler = _fail_fast(sender_node, "socket_can_sender")
+    receiver_fail_fast_handler = _fail_fast(receiver_node, "socket_can_receiver")
+
     return LaunchDescription([
         # Arguments
         interface_arg,
@@ -165,17 +172,15 @@ def generate_launch_description() -> LaunchDescription:
         # Validate CAN interface presence
         OpaqueFunction(function=_validate_can_interface),
 
+        # Register lifecycle handlers before nodes start to avoid races.
+        sender_configure_handler,
+        receiver_configure_handler,
+        sender_activate_handler,
+        receiver_activate_handler,
+        sender_fail_fast_handler,
+        receiver_fail_fast_handler,
+
         # Nodes
         sender_node,
         receiver_node,
-
-        # Handle lifecycle
-        _auto_configure(sender_node, "socket_can_sender"),
-        _auto_configure(receiver_node, "socket_can_receiver"),
-        _auto_activate(sender_node, "socket_can_sender"),
-        _auto_activate(receiver_node, "socket_can_receiver"),
-
-        # Fail fast if either node errors out
-        _fail_fast(sender_node, "socket_can_sender"),
-        _fail_fast(receiver_node, "socket_can_receiver"),
     ])
```

src/indomitus_rover_bringup/can_launch.md
OShevchuk-UCU, line 1: Docs mention entrypoint.sh and setup_can.sh, while implementation uses entrypoint.bash with auto CAN setup. Please align documentation with actual runtime flow and filenames.

docker/docker-compose.example.yml
OShevchuk-UCU, line 1: network_mode: host + NET_ADMIN/NET_RAW is correct for CAN, but please add a short note in docs that this is Linux-only and required for host CAN access.

OShevchuk-UCU (changes requested):
A few lifecycle, namespace, and runtime consistency issues still need cleanup to make the bring-up deterministic and maintainable on real hardware.

docker/Dockerfile
BohdanZasymovych: Redundant tabulation removed. In addition line moved after the commented out packages for installation

docker/entrypoint.bash
BohdanZasymovych: Installation of the sudo package and the addition to the sudoers in the Dockerfile removed. Sudo in the entrypoint removed. Any usage of the sudo is redundant because the default user is the root.

src/indomitus_rover_bringup/launch/can.launch.py
BohdanZasymovych: Race condition fixed, the event handlers now registered before the nodes

src/indomitus_rover_bringup/can_launch.md
BohdanZasymovych: The docstring in the launchfile changed, mentioning of the `can_setup.sh` script removed. The redundant now `can_setup.sh` script deleted

docker/docker-compose.example.yml
BohdanZasymovych: Note added

BohdanZasymovych:
@OShevchuk-UCU 
The requested changes were made. Please review them and approve the merge if everything is okay

OShevchuk-UCU (approved):
LGTM!

BohdanZasymovych:
Node tested on the real hardware with the two USB to CAN converters. Both receiving and transmitting works correctly.

PR #420, opened 2026-05-19
Feature/navigation/joystick connection

yagodanr:
## Description

Joystick control of the rover.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [x] Configuration change

## Changes Made

- Installed nodes /joy and /teleop_twist_joy to Dockerfile
- configured compose to forward input devices
- Added launch and config for joystick pipeline

## Testing (Future feature!!!)

- [x] All existing tests pass

### How to Test

1. enter docker container
2. launch
  ```bash
  ros2 launch indomitus_rover_bringup joy.launch.py
  ```
3. echo topics(2 other terminals):
  ```bash
  ros2 topic echo /joy
  ```
  ```bash
  ros2 topic echo /cmd_vel
  ```


docker/docker-compose.example.yml
```diff
@@ -19,7 +19,8 @@ services:
       - ROS_DOMAIN_ID=42
       - ROS_LOCALHOST_ONLY=0
       - CAN_INTERFACE=can0
-      - CAN_BITRATE=1000000
+    devices:
+      - /dev/input:/dev/input\
```
OShevchuk-UCU: There is a trailing backslash in the device mount path. This will mount `/dev/input\` instead of `/dev/input` or make the compose configuration invalid. It should be `/dev/input:/dev/input`.

OShevchuk-UCU (changes requested):
Looks good overall. I only found one critical issue in the compose device mapping path.

OShevchuk-UCU (approved):

docker/docker-compose.example.yml
BohdanZasymovych: Trailing backslash removed

PR #431, opened 2026-05-25
Feature/navigation/joystick interpreter

yuriifito:
## Description
We faced a problem that we should adapt joystick control to swerve kinematics. And also we had a need to be able to control other hardware mounted on rover via buttons on joystick. We implemented this functionality on this branch

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- add package with joystick_interpreter_node
- Implemented functionality:
    - control x and y speed via left stick
    - control angular movement via right stick horizontal axis
    - add enable/disable  service and corresponding button
    - add compact mode service to fold wheels inward (reduces rover footprint for transport)
    - update joysticks configs

## Testing

- [ ] All existing tests pass

### How to Test

1. Connect joystick to computer/laptop
2. Launch a rover, (via rover.launch.py). Or it can be tested in simulation (sim_gz_urdf.launch.py)
3. launch joystick's nodes (via joy.launch.py)
4. try to move rover using joystick


src/indomitus_rover_description/config/rover_motors.yaml
```diff
@@ -0,0 +1,4 @@
+rover_controller:
+  ros__parameters:
+    max_steer_deg: 90.0   # degrees
```
OShevchuk-UCU: Comment like
`maximum steering deviation from the forward direction`
would look better.

src/indomitus_rover_control/indomitus_rover_control/rover_kinematics_node.py
```diff
@@ -41,11 +41,16 @@
 
 from geometry_msgs.msg import Twist
 from indomitus_interfaces.msg import WheelTargets
+from std_srvs.srv import SetBool
 
 import rclpy
 from rclpy.node import Node
 
 
+_COMPACT_OFFSETS = {'FL': +math.pi, 'FR': -math.pi,
```
OShevchuk-UCU: ±math.pi compact offsets exceed the current ±90 steering limits

src/indomitus_rover_control/indomitus_rover_control/joystick_interpreter_node.py
```diff
@@ -0,0 +1,255 @@
...
+        return float(self.get_clock().now().nanoseconds) * 1e-9
+
+    def _toggle_motors(self):
+        if self._motor_toggle_pending:
+            self.get_logger().warn('Motor toggle request is already in flight')
+            return
+
+        target_enabled = not self._motors_enabled
+
+        if not self._motor_enable_client.wait_for_service(timeout_sec=1.0):
```
OShevchuk-UCU: Wait_for_service(1.0) blocks the single-threaded executor, freezing joystick processing and timeout handling for up to 1 s. Please use service_is_ready() instead.

src/indomitus_rover_control/indomitus_rover_control/joystick_interpreter_node.py
```diff
@@ -0,0 +1,255 @@
...
+        future.add_done_callback(
+            lambda completed_future, desired_state=target_enabled: self._on_motor_toggle_result(
+                completed_future,
+                desired_state,
+            )
+        )
+
+    def _toggle_compact_mode(self):
+        target = not self._compact_mode
+        if not self._compact_mode_client.wait_for_service(timeout_sec=1.0):
```
OShevchuk-UCU: Same blocking issue here — please use service_is_ready() instead of wait_for_service().

src/indomitus_rover_chassis_driver/src/chassis_driver_node.cpp
```diff
@@ -151,28 +209,39 @@ void ChassisDriverNode::sendDisableFrames() {
...
+    }
+
+    sendDisableFrames();
+    response->success = true;
+    response->message = "All chassis motors disabled";
 }
 
 // ── Command handling ──────────────────────────────────────────────────────────
 
 void ChassisDriverNode::onWheelTargets(const indomitus_interfaces::msg::WheelTargets::SharedPtr msg) {
```
OShevchuk-UCU, line 249: There is no stale-command watchdog in the chassis driver. If /wheel_targets stops arriving, steer motors may continue holding the last commanded angle indefinitely. Please add a timeout-based failsafe.

src/indomitus_rover_chassis_driver/src/chassis_driver_node.cpp
```diff
@@ -151,28 +209,39 @@ void ChassisDriverNode::sendDisableFrames() {
...
     if (!motors_enabled_) return;
 
-    const float angles[4] = {msg->fl_angle, msg->fr_angle, msg->rl_angle, msg->rr_angle};
-    const float speeds[4] = {msg->fl_speed, msg->fr_speed, msg->rl_speed, msg->rr_speed};
+    const float raw_angles[4] = {msg->fl_angle, msg->fr_angle, msg->rl_angle, msg->rr_angle};
+    const float speeds[4]     = {-msg->fl_speed, msg->fr_speed, -msg->rl_speed, msg->rr_speed};
```
OShevchuk-UCU, line 255: The {-1, +1, -1, +1} inversion is tightly coupled to motor mounting direction and drive_ids ordering. Consider moving this to a configurable parameter so it stays consistent if IDs or wheel layout change.

src/indomitus_rover_control/indomitus_rover_control/rover_kinematics_node.py
```diff
@@ -250,8 +294,9 @@ def _step_angle(self, current: float, target: float, dt: float) -> float:
         Difference is normalised to (-π, π) to take the shortest path.
         """
         diff = target - current
-        diff = (diff + math.pi) % (2 * math.pi) - math.pi  # shortest-path wrap
+        # diff = (diff + math.pi) % (2 * math.pi) - math.pi  # shortest-path wrap
         step = self._clamp(diff, -self.max_steer_rate * dt, self.max_steer_rate * dt)
+        # return self._clamp(current + step, -self.max_steer, self.max_steer)
         return current + step
```
OShevchuk-UCU, line 330: The final steering clamp is disabled, so current_angles can grow beyond the intended steering limits. Please restore a final clamp before publishing WheelTargets instead of relying only on the chassis driver clamp.

OShevchuk-UCU (changes requested):
Overall this is a solid improvement to the rover control stack. I left a few comments mainly around steering limits, stale-command handling, and blocking calls that could become problematic on real hardware.

src/indomitus_rover_control/indomitus_rover_control/rover_kinematics_node.py
yuriifito: we removed these limits. we empirically measured that our wheels are able to rotate up to 180 degrees in each direction

src/indomitus_rover_control/indomitus_rover_control/rover_kinematics_node.py
yuriifito: Steering clamp removed intentionally — wheel limits are asymmetric and side-dependent ([-π/2, 3π/2] vs [-3π/2, π/2]), so a single symmetric clamp here would incorrectly reject valid angles. Added a comment in code to prevent accidental revert.

src/indomitus_rover_description/config/rover_motors.yaml
yuriifito: I made an inaccuracy in matching the name and purpose (otherwise it would be too long). This is not a deviation from forward — it's a deviation from the reference angle defined by the current drive mode (e.g. crab, point-turn, etc.). Updated the comment to reflect that.

src/indomitus_rover_chassis_driver/src/chassis_driver_node.cpp
yuriifito: The inversion is a physical property of the chassis geometry — left wheels are mirrored relative to right wheels regardless of motor IDs or layout. Making it configurable would move a mechanical constant into YAML where it looks like a tunable parameter, and creates a risk of accidentally flipping a sign. If IDs change, only the FL/FR/RL/RR mapping changes, not the mounting direction.

PR #439, opened 2026-06-01
Refactor/shared/ros2 dependencies

BohdanZasymovych:
## Description
The PR provides the refactored handling of the dependencies. All dependencies for which it is possible are moved to the package.xml files instead of placing them in the Dockerfile.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

* The RViz launch file and configuration were moved from the `bringup` package to the new `viz` package. Please let me know if you can suggest a better place for these files. Maybe `sim` package or they should be in the separate package?
* The `package.xml` files for the `bringup` and `sim` packages were updated to include all required dependencies.
* The installation of `package.xml` dependencies in the `Dockerfile` was changed from using `apt` to `rosdep`.
* The `colcon build` and `rosdep install` commands were removed from `entrypoint.sh`.

## Testing

Docker image builds locally


docker/Dockerfile
```diff
@@ -50,15 +34,24 @@ FROM base AS dev
...
 
+COPY src/ /tmp/src_packages/src/
+RUN mkdir -p ${WS}/src && \
+    if [ "$INSTALL_SIM_TOOLS" = "true" ]; then \
+        cp -r /tmp/src_packages/src/* ${WS}/src/; \
+    else \
+        cd /tmp/src_packages/src && \
+        for pkg in *; do \
+            case "$pkg" in \
+                indomitus_rover_sim|indomitus_rover_viz) ;; \
```
OShevchuk-UCU: Potential issue: the `case` pattern uses `indomitus_rover_sim` and `indomitus_rover_viz`, but `src/` contains `rover_sim` and `rover_viz` (no matching directory names). This exclusion never matches, so sim/viz are still copied when `INSTALL_SIM_TOOLS=false`. Change the pattern to `rover_sim|rover_viz`.

docker/Dockerfile
```diff
@@ -68,22 +61,21 @@ CMD ["bash"]
...
+COPY src/ /tmp/src_packages/src/
 RUN mkdir -p ${WS}/src && \
-    cd /tmp/src_packages && \
-    find . -name "package.xml" -exec cp --parents {} ${WS}/src/ \; && \
-    cd ${WS} && \
-    rm -rf /tmp/src_packages
+    cd /tmp/src_packages/src && \
+    for pkg in *; do \
+        case "$pkg" in \
+            indomitus_rover_sim|indomitus_rover_viz) ;; \
```
OShevchuk-UCU: Same mismatch as the dev stage: `indomitus_rover_sim|indomitus_rover_viz` does not match `src/rover_sim` or `src/rover_viz`, so prod always copies simulation and visualization packages. Use `rover_sim|rover_viz` here as well.

src/rover_sim/package.xml
```diff
@@ -16,14 +16,18 @@
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
 
-  <depend>joint_state_publisher</depend>
+  <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>launch</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
+  <exec_depend>controller_manager</exec_depend>
+  <exec_depend>indomitus_rover_bringup</exec_depend>
```
OShevchuk-UCU: `indomitus_rover_bringup` is not a ROS package in this workspace (src/rover_bringup/package.xml declares `<name>rover_bringup</name>`; `src/indomitus_rover_bringup/` has no `package.xml`). `rosdep` cannot resolve this key and Docker builds may fail. Replace with `rover_bringup`

src/rover_sim/package.xml
```diff
@@ -16,14 +16,18 @@
...
+  <exec_depend>indomitus_rover_description</exec_depend>
```
OShevchuk-UCU: Same issue.  Replace with `rover_description`.

docker/Dockerfile
```diff
@@ -50,15 +34,24 @@ FROM base AS dev
...
+        cd /tmp/src_packages/src && \
+        for pkg in *; do \
+            case "$pkg" in \
+                indomitus_rover_sim|indomitus_rover_viz) ;; \
+                *) cp -r "$pkg" ${WS}/src/ ;; \
+            esac; \
+        done; \
+    fi
+
+RUN apt-get update && rosdep install -y -r -i --from-paths src --rosdistro ${ROS_DISTRO} && rm -rf /var/lib/apt/lists/*
```
OShevchuk-UCU: This runs `rosdep install -y -r -i --from-paths src` without `--ignore-src`, while CI (.github/workflows/build.yaml line 25) uses `rosdep install --from-paths src --ignore-src -r -y`. Without `--ignore-src`, workspace packages like `rover_description` may be treated as external dependencies and `rosdep` can fail or diverge from CI. Align the Docker command with CI.

docker/Dockerfile
```diff
@@ -68,22 +61,21 @@ CMD ["bash"]
...
-    rm -rf /tmp/src_packages
+    cd /tmp/src_packages/src && \
+    for pkg in *; do \
+        case "$pkg" in \
+            indomitus_rover_sim|indomitus_rover_viz) ;; \
+            *) cp -r "$pkg" ${WS}/src/ ;; \
+        esac; \
+    done
 
 RUN apt-get update && rosdep install -y -r -i --from-paths src --rosdistro ${ROS_DISTRO} && rm -rf /var/lib/apt/lists/*
```
OShevchuk-UCU: Prod-stage `rosdep`also omits `--ignore-src`, unlike CI. Combined with wrong `rover_sim/rover_viz` exclusions and invalid `indomitus_rover_*` keys in `rover_sim/package.xml`, this step is likely to fail or install unintended packages. Use the same flags as CI: `--from-paths src --ignore-src -r -y`.

OShevchuk-UCU (changes requested):
### Additional observations

While reviewing the repository, I noticed a few more issues that may be worth addressing:

1. **Simulation launch file references a non-existent xacro file**

   `sim_gz_urdf.launch.py` appears to load:

   ```text
   urdf/rover_gazebo.urdf.xacro
   ```

   while the repository contains:

   ```text
   urdf/indomitus_rover_gazebo.urdf.xacro
   ```

   If this is not intentional, the simulation launch may fail at runtime with a "file not found" error. It would be good to verify that the launch file and installed xacro filename are consistent.

2. **QUICKSTART documentation no longer matches container behavior**

   The documentation still states that the workspace is automatically built when the container starts. However, the build logic has been removed from `entrypoint.bash`, which now only configures CAN and sources ROS/install environments.

   As a result, new developers may expect the workspace to be built automatically while a manual `colcon build` is now required. The documentation should be updated accordingly.

3. **Missing message package dependencies in rover_control**

   `rover_control` imports messages from `geometry_msgs` and `sensor_msgs`, but these packages do not appear to be declared in `package.xml`.

   While this may work on systems where those packages are already installed through transitive dependencies, explicit declarations would make the package more self-contained and avoid issues on clean environments:

   ```xml
   <depend>geometry_msgs</depend>
   <depend>sensor_msgs</depend>
   ```

docs/QUICKSTART.md
```diff
@@ -1,3 +1,6 @@
+Here is the updated `QUICKSTART.md`. I have updated the "Build and start Docker" section to reflect that the workspace build is now a manual process.
+
+```markdown
```
OShevchuk-UCU: These should be removed, otherwise the document will render incorrectly and look unpolished.

OShevchuk-UCU (commented):
One more thing worth checking: after renaming `indomitus_rover_s1.urdf.xacro` to `rover_s1.urdf.xacro`, please make sure that all launch files, configs, documentation, and scripts have been updated and that there are no remaining references to the old filename.

Other than that, I don't see anything critical in this diff. The Docker/CI changes look reasonable, moving `RViz` into a dedicated `rover_viz package` makes sense, and the `colcon_defaults.yaml` addition looks fine.

yuriifito (approved):

PR #445, opened 2026-06-11
Feature/suspension/odometry

yuriifito:
## Description
Add rover_odometry_node that publishes msgs into '/odom' topic.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- add odometry node, dependencies for it
- change package description

## Testing

- [x] All existing tests pass

Tested it on real rover. The error accumulation rate in raw odometry is very low.

### DIFF
--- a/src/rover_chassis_driver/config/chassis_driver.yaml
+++ b/src/rover_chassis_driver/config/chassis_driver.yaml
@@ -7,7 +7,7 @@ chassis_driver:
 
     # Damiao MIT feedback fixed-point decoding ranges (must match motor registers 21/22/23)
     drive_pmax: 12.5    # rad  — register 21 (PMAX)
-    drive_vmax: 50.0    # rad/s — register 22 (VMAX)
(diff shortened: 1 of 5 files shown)


OShevchuk-UCU (commented):
The changes look reasonable to me.

One thing I'd like to clarify before merging: was the new odometry node actually tested as part of the full rover bringup (including `/odom` and TF publication), or only locally in isolation?

Also, could you confirm that the `drive_tmax` difference between the C++ default (200.0) and the YAML value (20.0) is intentional?

yuriifito:
1. Yes, we ran several tests on real rover with odometry node. It appeared to be precise, and accumulation of error was smth like 0.5-1 m, when rover travelled 15 meters and rotated several times in different points.
2. Somehow missed that yaml file, this value should be change there too, as it's a constant from documentation"

OShevchuk-UCU (approved):

PR #452, opened 2026-06-18
Refactor/suspension/ros2 control architecture

yuriifito:
## Description
A global refactoring of motor controls, rewrote via ros2 control.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [x] Configuration change

## Changes Made

- Instead of pipeline of 3 independent nodes, refactored them as ros2_control controller and hardware interface
- add option to launch docker container and nodes inside of it automatically on remote jetson
- add "granny" mode - that decreases speed by multiplying it by 0.1
- 

## Testing

Tested it on real rover, have video of it working correctly.

- [x] All existing tests pass


src/rover_hardware_interface/src/rover_hardware_interface.cpp
```diff
@@ -0,0 +1,704 @@
...
+    return hardware_interface::CallbackReturn::SUCCESS;
+}
+
+
+hardware_interface::CallbackReturn
+RoverHardwareInterface::on_deactivate(const rclcpp_lifecycle::State & /*prev*/)
+{
+    send_shutdown_frames();
+
+    hw_executor_->cancel();
```
OShevchuk-UCU: Executor lifecycle — on_deactivate() destroys the executor; after a subsequent OFF→ON cycle, polling, the watchdog, and services are no longer restored.
Suggested fix: Keep the executor alive across activate/deactivate cycles. on_deactivate() should only close the socket and disable the motors. If the executor can be destroyed, recreate it in on_configure() or lazily during on_activate().

src/rover_hardware_interface/src/rover_hardware_interface.cpp
```diff
@@ -0,0 +1,704 @@
...
+    uint32_t id, const uint8_t * data, uint8_t dlc, bool is_extended)
+{
+    if (can_fd_ < 0) return false;
+
+    struct can_frame frame{};
+    frame.can_id  = is_extended ? (id | CAN_EFF_FLAG) : (id & CAN_SFF_MASK);
+    frame.can_dlc = dlc;
+    std::memcpy(frame.data, data, dlc);
+
+    const ssize_t nbytes = ::write(can_fd_, &frame, sizeof(frame));
```
OShevchuk-UCU, line 440: `send_can_frame()` is called from two different threads without synchronization: from `write()` (the `controller_manager` control loop at ~100 Hz) and from timers running on `hw_node_thread_` (1 Hz status polling and the watchdog). Both write to the same `can_fd_` using `::write()`. This creates a data race and an unpredictable frame ordering on the CAN bus (poll/watchdog frames may interleave with steer/drive frames within a single control cycle). RX already uses `feedback_mutex_`; TX should similarly introduce a `can_tx_mutex_` and lock it inside `send_can_frame()` (or route all TX through a single dedicated thread).

src/rover_hardware_interface/src/rover_hardware_interface.cpp
```diff
@@ -0,0 +1,704 @@
...
+
+    // ── 1 Hz status poll (Steadywin 0xAE + 0xA3, Damiao reg 80) ──────────────
+    status_poll_timer_ = node->create_wall_timer(1s, [this]() {
+        if (!motors_enabled_) return;
+        for (std::size_t i = 0; i < NUM_WHEELS; ++i) {
+            auto f1 = steadywin_protocol::buildStatusQueryFrame(steer_ids_[i]);
+            auto f2 = steadywin_protocol::buildAbsAngleQueryFrame(steer_ids_[i]);
+            auto f3 = damiao_protocol::buildReadRegisterFrame(drive_ids_[i], 80);
+            send_can_frame(f1.id, f1.data.data(), f1.dlc);
+            send_can_frame(f2.id, f2.data.data(), f2.dlc);
```
OShevchuk-UCU, line 197: These `send_can_frame()` calls run on `hw_node_thread_`, while `write()` sends commands from the controller_manager thread. Both use the same `can_fd_`, so a shared mutex is required inside `send_can_frame()`.

src/rover_chassis_controller/src/swerve_controller.cpp
```diff
@@ -0,0 +1,555 @@
...
+    vx_smoothed_ = limiters_[0].update(target_vx_, dt);
+    vy_smoothed_ = limiters_[1].update(target_vy_, dt);
+    wz_smoothed_ = limiters_[2].update(target_wz_, dt);
+
+    const double vx = vx_smoothed_;
+    const double vy = vy_smoothed_;
+    const double wz = wz_smoothed_;
+
+    // // Step 2: Read current hardware steering positions
+    // read_current_angles(); --- IGNORE --- (already read in on_activate and updated in step_angle)
```
OShevchuk-UCU, line 183: The `IGNORE` comment explains the intention (seed once in `on_activate()`, then rely on `step_angle()`), but `step_angle()` updates an internal model based on the commanded position, not on encoder feedback. `current_angles_` is later used by TRANSIT, IK, and the `cos(work - current)` alignment logic, so if the steering motor lags behind, the controller is unaware of it.

If open-loop steering is an intentional design choice, the assumption that "the steering actuator always follows the commanded position" should be explicitly documented. Otherwise, `read_current_angles()` should be called at the beginning of `update()` (or encoder feedback should be synchronized into `current_angles_` before the alignment calculations).

src/rover_bringup/package.xml
OShevchuk-UCU, line 1: `rover_bringup/package.xml` does not declare `controller_manager`, `robot_state_publisher`, `rover_chassis_controller`, or `joint_state_broadcaster`. This may break `rosdep` or a clean install on a fresh machine.

OShevchuk-UCU (changes requested):
Solid refactor toward `ros2_control` — tested on the real rover. **Request changes** before merge.

### Blocker

- **Auto-enable on activation**  — calls `send_enable_frames()` immediately when hardware goes ACTIVE. Conflicts with README ("disabled by default") and may bypass explicit operator enable unless intentional.

### Non-blocking follow-up

- **`test_kinematics.py`** — imports removed `rover_kinematics_node`; fails `colcon test` (not CI today — pipeline only runs `colcon build`). Rewrite against `SwerveKinematics` or remove.
- **Dual CAN on `can0`** — chassis via raw SocketCAN, peripherals via `ros2_socketcan`. Not a bug if intentional; document in README.
- **Docs drift** — `QUICKSTART.md`, hardware README, commented-out `autolaunch` in prod compose.

Inline comments cover executor lifecycle, CAN TX mutex, open-loop steering, and missing `exec_depend`.

src/rover_hardware_interface/src/rover_hardware_interface.cpp
yuriifito: moved executor destroyment code to on_cleanup and on_shutdown just in case

src/rover_hardware_interface/src/rover_hardware_interface.cpp
yuriifito: fixed by mutex added to solve previous issue

src/rover_chassis_controller/src/swerve_controller.cpp
yuriifito: We know about this issue. It's difficult to fix it here without knowing odometry final c++ architecture and read procedure. I will fix it in new odometry PR. Currently it works well, as motors have enough time to turn

OShevchuk-UCU (approved):
LGTM after review fixes; please update README lifecycle wording in a follow-up.

PR #456, opened 2026-06-22
Feature/navigation/zed2i integration

BohdanZasymovych:
## Description

PR provides docker configuration, launch files and updated docker setup needed for the work with ZED 2i stereo camera

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- launch file for the stereo camera node added
- config for the stereo camera node added 
- hw-builder stage added to the Dockerfile
- other minor changes and fixes

## Testing

I have verified work of the stereo camera in both scenarios: with my laptop and the Jetson. Launch file successfully launches the node and it publishes what it needs to


docker/Dockerfile
```diff
@@ -1,6 +1,10 @@
 ARG UBUNTU_VERSION=22.04
-ARG ROS_DISTRO=humble
+ARG TARGET_ROS_DISTRO=humble
```
OShevchuk-UCU, line 2: This PR renames ROS_DISTRO to TARGET_ROS_DISTRO, but the prod-stage RUN on line 198 still sources ${ROS_DISTRO}. That variable is no longer defined, so prod colcon build fails or runs without ROS sourced. Update line 198 to use TARGET_ROS_DISTRO.

docker/Dockerfile
```diff
@@ -1,6 +1,10 @@
...
+ARG BASE_IMAGE_HW_BUILDER
```
OShevchuk-UCU: BASE_IMAGE_HW_BUILDER and BASE_IMAGE have no defaults, but FROM on lines 30 and 99 require them. A plain `docker build` without build-args fails with an invalid base image reference. Add defaults matching docker-compose/CI, or fail early when these args are unset.

docker/Dockerfile
```diff
@@ -10,22 +14,110 @@ FROM ubuntu:${UBUNTU_VERSION} AS cacher
...
         for pkg in ${SIMULATION_PKGS}; do find . -name "$pkg" -type d -exec rm -rf {} + ; done; \
     fi && \
     mkdir -p /extracted_xmls && \
     find . -name "package.xml" -exec cp --parents {} /extracted_xmls/ \;
 
 
+####################
+# HW-BUILDER
+####################
+FROM ${BASE_IMAGE_HW_BUILDER} AS hw-builder
```
OShevchuk-UCU, line 30: BASE_IMAGE_HW_BUILDER has no default in the Dockerfile (line 4). If this build-arg is not passed, this FROM fails with an invalid image reference. Add a default or require it explicitly in every build path.

docker/Dockerfile
```diff
@@ -10,22 +14,110 @@ FROM ubuntu:${UBUNTU_VERSION} AS cacher
...
+    -DCMAKE_EXE_LINKER_FLAGS='-Wl,--allow-shlib-undefined'" \
+    && rm -rf build/ log/ \
+    && rm -rf /tmp/fake_gpu
+
+
 ####################
 # BASE
 ####################
-FROM ros:${ROS_DISTRO}-ros-base AS base
+FROM ${BASE_IMAGE} AS base
```
OShevchuk-UCU, line 110: BASE_IMAGE has no default in the Dockerfile (line 5). If this build-arg is not passed, this FROM fails with an invalid image reference. Add a default matching docker-compose/CI or fail early when unset.

docker/Dockerfile
OShevchuk-UCU: This still sources ${ROS_DISTRO}, but the Dockerfile now defines TARGET_ROS_DISTRO only (line 2). ROS_DISTRO is unset here, so prod colcon build runs without a sourced ROS environment and the image build fails or produces a broken workspace. Use: source /opt/ros/${TARGET_ROS_DISTRO}/setup.bash

docker/docker-compose.prod.yaml
```diff
@@ -1,15 +1,19 @@
...
   rover_prod:
+    platform: linux/arm64
+    runtime: nvidia
     build:
       context: .
       dockerfile: docker/Dockerfile
       target: prod
       args:
-        ROS_DISTRO: humble
+        BASE_IMAGE_HW_BUILDER: ${BASE_IMAGE_HW_BUILDER:-stereolabs/zed:5.4-devel-l4t-r36.4}
```
OShevchuk-UCU, line 10: These build-args supply BASE_IMAGE_HW_BUILDER and BASE_IMAGE because the Dockerfile has no defaults for them. Without these values, docker compose build fails at the FROM stage.

docker/docker-compose.prod.yaml
OShevchuk-UCU: This healthcheck sources /opt/ros/${ROS_DISTRO}, but the PR replaces ROS_DISTRO with TARGET_ROS_DISTRO in build args and Dockerfile ENV. ROS_DISTRO is not set in the environment section below, so the healthcheck can fail with an empty distro path. Use TARGET_ROS_DISTRO or set ROS_DISTRO=${TARGET_ROS_DISTRO} in environment.

docker/docker-compose.dev.example.yaml
```diff
@@ -1,16 +1,21 @@
 services:
   rover_dev:
+    platform: linux/amd64
     build:
       context: .
       dockerfile: docker/Dockerfile
       target: dev
       args:
-        ROS_DISTRO: humble
+        BASE_IMAGE_HW_BUILDER: ${BASE_IMAGE_HW_BUILDER:-stereolabs/zed:5.4-devel-cuda12.8-ubuntu22.04}
```
OShevchuk-UCU, line 9: These build-args supply BASE_IMAGE_HW_BUILDER and BASE_IMAGE because the Dockerfile has no defaults for them. Without these values, docker compose build fails at the FROM stage.

docker/docker-compose.dev.example.yaml
OShevchuk-UCU: This healthcheck sources /opt/ros/${ROS_DISTRO}, but the PR replaces ROS_DISTRO with TARGET_ROS_DISTRO in build args and Dockerfile ENV. ROS_DISTRO is not set in the environment section below, so the healthcheck can fail with an empty distro path. Use TARGET_ROS_DISTRO or set ROS_DISTRO=${TARGET_ROS_DISTRO} in environment.

src/rover_sensors/launch/zed2i.launch.py
```diff
@@ -0,0 +1,37 @@
...
+from launch_ros.substitutions import FindPackageShare
+from launch.substitutions import PathJoinSubstitution, LaunchConfiguration
+from launch.actions import DeclareLaunchArgument
+
+DEFAULT_ZED2I_CONFIG_NAME = "zed2i.yaml"
+CAMERA_MODEL = "zed2i"
+
+
+def generate_launch_description():
+    child_launch_file_path = PathJoinSubstitution([
```
OShevchuk-UCU, line 13: This resolves zed_wrapper at runtime, but rover_sensors/package.xml does not declare exec_depend on zed_wrapper. rosdep install --from-paths src will not pull it, and ros2 launch fails without manually sourcing /opt/hw_ws/install/setup.bash. Add <exec_depend>zed_wrapper</exec_depend> to package.xml.

.github/workflows/pr_pipeline.yaml
```diff
@@ -92,6 +92,8 @@ jobs:
           push: true
           build-args: |
             INSTALL_SIM_TOOLS=true
+            BASE_IMAGE_HW_BUILDER=stereolabs/zed:5.4-devel-cuda12.8-ubuntu22.04
+            BASE_IMAGE=stereolabs/zed:5.4-runtime-cuda12.8-ubuntu22.04
```
OShevchuk-UCU, line 118: These build-args are required because Dockerfile ARG BASE_IMAGE_HW_BUILDER and BASE_IMAGE have no defaults. Without them, docker build fails. Consider adding matching defaults in the Dockerfile so local builds do not depend only on CI/compose.

.github/workflows/publish_image.yaml
```diff
@@ -68,6 +68,9 @@ jobs:
           target: prod 
           platforms: linux/arm64
           push: true
+          build-args: |
+            BASE_IMAGE_HW_BUILDER=stereolabs/zed:5.4-devel-l4t-r36.4
+            BASE_IMAGE=stereolabs/zed:5.4-devel-l4t-r36.4
```
OShevchuk-UCU, line 84: BASE_IMAGE_HW_BUILDER and BASE_IMAGE are required because the Dockerfile declares them without defaults. CI must always pass these args or the build fails at the FROM stage.

.github/workflows/publish_image.yaml
```diff
@@ -121,6 +124,8 @@ jobs:
           push: true
           build-args: |
             INSTALL_SIM_TOOLS=true
+            BASE_IMAGE_HW_BUILDER=stereolabs/zed:5.4-devel-cuda12.8-ubuntu22.04
+            BASE_IMAGE=stereolabs/zed:5.4-runtime-cuda12.8-ubuntu22.04
```
OShevchuk-UCU, line 164: BASE_IMAGE_HW_BUILDER and BASE_IMAGE are required because the Dockerfile declares them without defaults. CI must always pass these args or the build fails at the FROM stage.

OShevchuk-UCU (changes requested):
Thanks for the ZED2i bringup — the thin launch wrapper around `zed_camera.launch.py`, pinned `hw-builder` commits, and standardized Docker `hw-builder` stage for both LiDAR and ZED are solid improvements. Manual validation on both a laptop and Jetson is also appreciated.

## What looks good

- Clean `zed2i.launch.py` delegating to the upstream `zed_camera.launch.py` with `camera_model` and `ros_params_override_path`.
- `hw-builder` stage uses pinned SHAs, shallow Git fetch, submodule initialization, and fake CUDA stubs to support GPU-less CI builds.
- Production Docker Compose correctly enables `runtime: nvidia`, mounts ZED settings/resources, and targets the `linux/arm64` platform.
- Renaming the RPLidar launch file to `rplidar_s2.launch.py` aligns with the repository naming convention, and the README table has been updated accordingly.

---

# Minor / Follow-up

## Documentation

- `rplidar.md` still references `rplidar_s2_launch.py` and the command:
  ```bash
  ros2 launch rover_sensors rplidar_s2_launch.py
  ```
  These should be updated to `rplidar_s2.launch.py`.

- There is currently no `zed2i.md` equivalent to `rplidar.md`. It would be useful to document:
  - sourcing `/opt/hw_ws/install/setup.bash`
  - launch commands
  - expected topics and TF frames
  - USB/container requirements

---

## Docker / Deployment

- The production ARM64 workflow uses `stereolabs/zed:5.4-devel-l4t-r36.4` for both the builder and runtime images, while the AMD64 workflow already uses a runtime image for `BASE_IMAGE`. If the SDK is only required during build, consider switching the production runtime image to the corresponding runtime variant to reduce image size.

- The cleanup command:
  ```bash
  find src -name ".github" -type f -delete
  ```
  does not remove `.github` directories because they are directories rather than files.

- The Docker Compose healthcheck uses `$$ROS_DISTRO`, while the Dockerfile now uses `TARGET_ROS_DISTRO`. It would be worth confirming that the Stereolabs base image always exports `ROS_DISTRO`, or aligning the healthcheck with `TARGET_ROS_DISTRO`.

---

## Configuration (`zed2i.yaml`)

- The configuration file is effectively a full upstream parameter dump (~219 lines). Consider keeping only rover-specific overrides to simplify future upgrades and reviews.

- The default settings are relatively aggressive for embedded deployment (`grab_frame_rate: 60`, `NEURAL_LIGHT` depth, point cloud publishing, positional tracking enabled). It would be useful to profile these settings on Jetson under the full rover workload and tune them if necessary.

---

## Launch File

- `zed2i.launch.py` currently exposes only `config_path`. Consider also forwarding commonly used launch arguments such as:
  - `serial_number`
  - `namespace`

  This would simplify multi-camera deployments and explicit device selection.

---

## CI

- There is currently no automated validation that:
  - `zed2i.launch.py` resolves correctly;
  - `zed2i.yaml` is installed into the package.

  A lightweight CI smoke test (for example using `ros2 launch ... --show-args`) would help detect packaging regressions.

---

# Questions for the Author

### TF publishing

`zed2i.yaml` enables:

- `publish_tf: true`
- `publish_map_tf: true`
- `publish_imu_tf: true`

Will the ZED node be used alongside Nav2/AMCL on the rover? If so, how is ownership of the `map → odom` transform handled to avoid multiple publishers? If the camera is currently intended only for image/depth streaming, documenting when TF publishing should be disabled would help future integrators.

### `hw_ws` sourcing

Is

```bash
source /opt/hw_ws/install/setup.bash
```

always required before launching the ZED node (similar to the LiDAR setup)? If so, it would be helpful to include this as part of the documented launch procedure.

---

# Recommendation

After addressing the blocking inline review comments, I think this PR will be in good shape.

The remaining items above are mostly documentation, deployment, configuration, and maintainability improvements that can be addressed either in this PR or in a follow-up.

BohdanZasymovych:
> The production ARM64 workflow uses stereolabs/zed:5.4-devel-l4t-r36.4 for both the builder and runtime images, while the AMD64 workflow already uses a runtime image for BASE_IMAGE. If the SDK is only required during build, consider switching the production runtime image to the corresponding runtime variant to reduce image size.

The develop image for the production container is chosen intentionally. For the x86 the develop and runtime images have big size difference of about 4 GB but for some reason there is no practical size difference between this images for ARM64. So by choosing the same images for both hw-builder and base we are allowing to download only one  image instead of two and reuse it between stages. This makes initial build without a cache faster and saves about 17 GB on the builder machine (size of the uncompressed image). 

image
image

BohdanZasymovych:
> The configuration file is effectively a full upstream parameter dump (~219 lines). Consider keeping only rover-specific overrides to simplify future upgrades and reviews.

Currently i have not researched all the configuration params available and don't know which we really need and which can be thrown away. That is why I am keeping the full config which was copied from the official repo with ros wrapper for ZED2i stereo camera. In the future I will create the smaller config with useful params.

BohdanZasymovych:
> Will the ZED node be used alongside Nav2/AMCL on the rover?

On the rover we will not use AMCL at all since it requires known in advance map, AMCL is just temporary solution to handle localization. Instead we will use SLAM and the SLAM node will be responsible for publishing `map->odom` transform. For now we are keeping the publish of this transform by the camera turned on since we do not have SLAM implemented yet.

BohdanZasymovych:
**Key changes:**
* Installation of the repo with zed_description package added to the dockerfile
* 3D model of the camera added to the simulation
* While we keeping the camera responsible for publishing both `map->odom` and `odom->zed2i_camera_link` the relations of the links changed to make zed2i_camera_link the root of the tree (required by the docs of the camera)
* ROS_DISTRO environmental variable replaced everywhere by the TARGET_ROS_DISTRO
* Default values for the base images in  the dockerfile added to make the local `docker build` with no args build the dev image

Thank you for the other suggestions, I have reviewed them and will address them in the following PRs.

OShevchuk-UCU (changes requested):
Looks good overall, thanks for addressing the blocking comments. 

I only noticed one small remaining issue: entrypoint.bash still sources ${WH_WS} instead of ${HW_WS}, which looks like a typo. Everything else can be handled in the planned follow-up PRs.

OShevchuk-UCU (approved):

PR #463, opened 2026-06-23
Feature/navigation/rplidar s2

yagodanr:
## Description
Added rplidar node and launch files to `rover_sensors` package

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Testing

- [ ] All existing tests pass

### How to Test

1. build
2. `source install/setup.bash`
3. `ros2 launch rover_sensors rplidar_s2_launch.py `

## Additional Notes

original package used `sllidar-***` naming, but I decided, that it would be more convenient to name public files(node and launch file) as 'rplidar-***', because that is the name of the sensor. Mentioned this in docs

When lidar is not connected raises: ` Error, unexpected error, code: 80008004`


BohdanZasymovych (changes requested):
Please completely remove the cloned external repository from our codebase. Currently, it's bloating our repository with source code that we don't actually need to modify.

To handle this more cleanly (as discussed in the ROS ERC Telegram group), we should clone the external repo containing the basic LiDAR nodes into a separate workspace (`/opt/hw_ws/`) directly inside the Dockerfile, and then build it there. This ensures all required packages are baked into the Docker image without cluttering our repo.

Also, keep in mind Oleksandr Shevchuk’s recommendation: if the package doesn't have a stable release branch or updates frequently, we should pin it to a specific commit hash. This will prevent our Docker builds from suddenly breaking due to unexpected upstream changes.

BohdanZasymovych:
I have updated Dockerfile to contain this additional workspace and merged changes to this branch.

yagodanr:
moved sdk into docker file to pull specific hash commit of needed implementation into `/opt/hw_ws`. Now `/opt/ws/src/rover_sensors` in CMakeLists.txt links not the `./sdk/`, but full path to the package in `/opt/hw_ws/src/sllidar_ros2/sdk/`

BohdanZasymovych (dismissed):

BohdanZasymovych (commented):
General things which i understand looks good to me.
@OShevchuk-UCU review please.

OShevchuk-UCU (commented):
#### Overall this looks good and I think the integration approach is solid. I only have a few documentation-related comments before merge.

---
**README / rplidar.md**

The documentation currently assumes that `sllidar_ros2` is available, but that is only true inside the project Docker image where `/opt/hw_ws` is sourced automatically. Outside Docker, the launch will fail with `package 'sllidar_ros2' not found`.

Could we explicitly document that the driver comes from `/opt/hw_ws` (or that the Docker image is the recommended runtime environment)?

---
**rplidar.md**

The documentation says the executable was renamed to `rplidar_*`, but the launch still starts `sllidar_ros2/sllidar_node`; only the ROS node name is `rplidar_node`.

I'd suggest rewording this section to avoid implying that the executable itself was renamed.

---
**rplidar_s2_launch.py**

The default `frame_id` is `laser`, but there is currently no corresponding TF/URDF frame in `nav-core`. That is fine for this PR, but it would be helpful to mention in the documentation that navigation will require a matching TF (e.g. `lidar_link`) or a static transform before Nav2 can consume `/scan`.

yagodanr:
@OShevchuk-UCU, I was asked to also add config file for default parameters, after which I would update documentation. Just pushed the changes. Please review. Sorry for miscommunication

BohdanZasymovych (approved):

PR #465, opened 2026-06-23
Refactor/suspension/ros2 control wheels odometry

yuriifito:
## Description
Rewrite odometry node as controller in c++

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- rename package rover_chassis_controller -> rover_controller
- add wheels odometry as controller in rover_controller package
- remove old odometry node in python from rover_control
- rename package rover_control -> rover_teleop (because now it caontains only joystick interpreter node)

## Testing

- [x] All existing tests pass


OShevchuk-UCU (commented):
Could you please do a quick repo-wide search for stale references after the package renames (`rover_control`, `rover_odometry_node`, `rover_kinematics_node`)?

Could you verify the behaviour when RoverHardware transitions from INACTIVE to ACTIVE? Since odometry_controller starts immediately, it would be good to confirm there is no encoder discontinuity causing an initial odom jump.

yuriifito:
1. the only mentions of these nodes are in docs that are rewritten in another PR

2. I had previously bug with initial jump in odometry. Now I fixed that problem, so wheel odometry initial position is 0

OShevchuk-UCU (approved):

PR #466, opened 2026-06-23
Feature/suspension/twist mux cherry pick

yuriifito:
## Description
Add `twist_mux` node to multiplex multiple `cmd_vel` sources with priority-based selection

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [x] Configuration change

## Changes Made
- Added `twist_mux` node to `rover.launch.py` with remapping of `/cmd_vel_out` -> `/cmd_vel`
- Added `twist_mux.yaml` config with three input topics: `cmd_vel_joy` (priority 100), `cmd_vel_nav` (priority 50), `cmd_vel_ext` (priority 10)

## Testing
- [x] All existing tests pass

### DIFF
--- a/src/rover_bringup/config/twist_mux.yaml
+++ b/src/rover_bringup/config/twist_mux.yaml
@@ -1,0 +1,25 @@
+# rover_bringup/config/twist_mux.yaml
+twist_mux:
+  ros__parameters:
+    topics:
+      joystick:
+        topic:   "cmd_vel_joy"
+        timeout: 0.5
+        priority: 100
+      
+      nav2:
+        topic:   "cmd_vel_nav"
+        timeout: 1.0
+        priority: 50
+
+      external:
+        topic:   "cmd_vel_ext"
+        timeout: 0.5
+        priority: 10 
+
+    # in a case of connection lost, we can publish 0 in this topic
+    locks:
+      e_stop:
+        topic:   "twist_mux/e_stop"
+        timeout: 0.2
+        priority: 255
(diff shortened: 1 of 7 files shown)


OShevchuk-UCU (commented):
The teleop remap does not match the configured twist_mux input.

`joy.launch.py` remaps joystick output to `/cmd_vel/teleop`, but `twist_mux.yaml` listens on `cmd_vel_joy` (`/cmd_vel_joy` in the root namespace). As a result, teleop commands will not reach twist_mux and therefore will not be forwarded to `/cmd_vel`.

Could you align these topic names, for example by remapping teleop to `/cmd_vel_joy` or by changing the mux input topic to `cmd_vel/teleop`?
___

Since `rover.launch.py` now starts a `twist_mux` node, please add `twist_mux` as a runtime dependency in `rover_bringup/package.xml`, for example:

`<exec_depend>twist_mux</exec_depend>`

Otherwise a clean install / rosdep-based setup may fail at runtime if `twist_mux` is not installed manually.

___

The e-stop lock topic looks like future wiring only, since this PR does not add any publisher for `twist_mux/e_stop`. Could you clarify this in the config/comment and document that the lock expects `std_msgs/Bool`?

yuriifito:
tested it, now it works fine

OShevchuk-UCU (approved):
Looks good to me.

One small suggestion: could you also update the PR description to match the final implementation? It still mentions the future e-stop lock, but that configuration is no longer part of the current PR.

PR #468, opened 2026-06-24
Feature/suspension/ros2 control sim

yuriifito:
## Description
Fix simulation to support new architecture and simulate odometry.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- add odometry to sim and change launch file for sim
- rename parameters -> config in sim package
- rename hardware_component GazeboSystem -> RoverHardware
- swerve drive now takes data from wheels encoders and takes it into account
- make rover movement response quicker

## Testing

- [x] All existing tests pass


OShevchuk-UCU:
Could you confirm that sim odometry direction was validated after moving `kDriveSigns` out of `odometry_controller` and into the hardware read path?

For hardware this looks like the right place to canonicalize encoder signs, but the sim path uses `gz_ros2_control/GazeboSimSystem`, so it bypasses `RoverHardwareInterface`. If Gazebo joint axes already match the expected convention, this is fine, but it would be good to explicitly verify forward / rotate `/odom` direction in sim.

OShevchuk-UCU:
In `rover_sim/config/controllers.yaml`, the swerve controller still uses `drive_joints` / `steering_joints`, while the controller appears to expect `drive_joint_names` / `steer_joint_names`.

Could you align the parameter names? The current setup may work only because the controller falls back to its internal defaults, which could become a maintenance issue later.

OShevchuk-UCU:
Could you clarify the PR description? It looks like encoder feedback is currently used only for alignment/transit logic, not for full closed-loop steering control.

yuriifito:
1. Yes, validated and confirmed in sim. Forward, rotate, and odom directions all work correctly.

The reason sim works without `kDriveSigns` is that `gz_ros2_control` relies entirely on the URDF joint axes. And there we described for each wheel direction of forward rotation.

yuriifito:
2. fixed

yuriifito:
3. To clarify: the controller doesn't control the steering motors using their position and PID. 

Instead, the controller reads the encoders as a safety check. It plans a smooth path for the wheels, and if the physical wheels can't keep up (lag behind), the controller automatically slows down the whole rover till is doesn't reach desired angle

OShevchuk-UCU (approved):

PR #470, opened 2026-06-24
Hotfix: pr pipeline fixed to correctly handle image tags

BohdanZasymovych:
## Description

PR pipeline fixed to sanitize image tags and use only existing main-dev and develop-dev images for the fallback


yuriifito (approved):
lgtm

PR #474, opened 2026-06-28
Feature/navigation/setup nav2

MunenNoKage:
## Description
Sets up Nav2. Configures the three core navigation servers (planner, controller, BT navigator), AMCL localisation, and a test map.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [x] Configuration change

## Changes Made
- Added `rover_navigation` package with `nav2_params.yaml`, `navigation.launch.py`, test map, and custom behaviour trees
- Configured MPPI controller for omni drive
- Added `NAVIGATION.md` documenting setup
- Added a test simulation world
- Updated `rover_viz` RViz configurations to visualize path planning, costmap layers
- Fixed a telemetry bug causing multiple joint state publishers to conflict
- Integrated a LiDAR sensor to provide laser scans (`/scan`) for costmap obstacle

## Testing

- [ ] All existing tests pass

### How to Test
 
1. Launch simulation:
```bash
ros2 launch rover_sim sim_gz_nav2.launch.py
```
 
2. In a second terminal, launch navigation:
```bash
ros2 launch rover_navigation navigation.launch.py
```
 
3. In a third terminal, launch RViz:
```bash
ros2 launch rover_viz rviz.launch.py use_sim:=true
```

## Additional Notes
Please note that AMCL localization is currently bundled straight into the rover_navigation launch and lifecycle pipeline for simplicity during testing. In a future development sprint, this architectural component must be decoupled and managed separately from the core path-planning servers. Crucial implementation notes and sections targeted for this future split have been explicitly marked with # NOTE

@BohdanZasymovych @OShevchuk-UCU


src/rover_description/urdf/ros2_control_sim.xacro
```diff
@@ -64,4 +64,18 @@
 
   </ros2_control>
 
+  <!-- Publishes /odom topic and odom to base_link transform required by nav2.
+      Bridged to ROS via /odom and /tf entries in bridge_gz.yaml. -->
+  <gazebo>
+    <plugin filename="gz-sim-odometry-publisher-system" name="gz::sim::systems::OdometryPublisher">
```
OShevchuk-UCU: This `OdometryPublisher` plugin is added to `rover_sim.urdf.xacro`, which is also used by the standard `sim_gz.launch.py`. That launch file still starts `odometry_controller`, while `bridge_gz.yaml` now bridges both `/odom` and `/tf` from Gazebo.

As a result, the default simulation will have **two odometry sources** publishing `/odom` and the `odom → base_link` transform, which can lead to conflicting TF data and unreliable odometry.

Please ensure only one odometry source is active. For example, enable the Gazebo odometry plugin only for the Nav2 launch (or another dedicated simulation mode), or disable `odometry_controller` whenever the Gazebo odometry plugin is used.

src/rover_navigation/package.xml
OShevchuk-UCU, line 1: `navigation.launch.py` starts `map_server`, `amcl`, and the MPPI controller (`nav2_mppi_controller::MPPIController` configured in `nav2_params.yaml`), but `package.xml` does not declare the corresponding runtime dependencies (`nav2_map_server`, `nav2_amcl`, and `nav2_mppi_controller`). At the same time, it declares `nav2_regulated_pure_pursuit_controller`, which is not used by this configuration.

As a result, a clean `rosdep install --from-paths src` will not install all packages required to run this launch file. Please add the missing `exec_depend` entries and remove the unused `nav2_regulated_pure_pursuit_controller` dependency.

src/rover_navigation/config/nav2_params.yaml
```diff
@@ -0,0 +1,354 @@
...
+        publish_voxel_map: true
+        origin_z: 0.0
+        z_resolution: 0.05
+        z_voxels: 16
+        max_obstacle_height: 2.0
+        mark_threshold: 0
+        observation_sources: scan
+        scan:
+          topic: /scan
+          sensor_frame: lidar_link
```
OShevchuk-UCU: Both costmap scan sources use `sensor_frame: lidar_link` (same issue at line 321), but on `nav-core` the RPLidar driver publishes scans with `frame_id: laser_link`, and the hardware URDF does not define `lidar_link`.

After merge, hardware `/scan` data will not be transformed into the costmaps correctly unless these frames match.

Please align the frame name across the URDF, `rplidar_s2.yaml`, and the Nav2 costmap parameters, or make `sensor_frame` configurable through a launch-time parameter.

src/rover_viz/rviz/robot.rviz
```diff
@@ -139,69 +75,169 @@ Visualization Manager:
...
+        History Policy: Keep Last
+        Reliability Policy: Reliable
+        Value: /particle_cloud
+      Value: true
+
   Enabled: true
   Global Options:
     Background Color: 48; 48; 48
-    Fixed Frame: base_link
+    Fixed Frame: map
```
OShevchuk-UCU: `Fixed Frame` is now set to `map`, but `rviz.launch.py` still defaults to `use_sim:=false` and does not start Nav2 or AMCL.

As a result, a plain `ros2 launch rover_viz rviz.launch.py` will fail the TF lookup because the `map` frame does not exist, and the robot model may not display. This is a regression from the previous `base_link` default.

Please keep `base_link` as the default RViz fixed frame, and switch to `map` only when navigation is running, for example via a `use_nav` launch argument or a separate navigation-specific RViz config.

src/rover_sensors/CMakeLists.txt
```diff
@@ -1,31 +1,46 @@
...
+if(NOT CMAKE_C_STANDARD)
+    set(CMAKE_C_STANDARD 99)
+endif()
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
OShevchuk-UCU: This branch reintroduces a skeleton `rover_sensors` package, while `nav-core` already contains a functional implementation with the RPLidar bringup.

Although Git will report merge conflicts only in `package.xml` and `CMakeLists.txt`, resolving those conflicts incorrectly (for example, by taking the skeleton version) can remove required dependencies such as `sllidar_ros2` and break the hardware bringup, even if the launch files themselves remain in the repository.

Please rebase onto the current `nav-core` and avoid modifying `rover_sensors` in this PR unless the changes are directly related to the navigation stack.

OShevchuk-UCU (changes requested):
The Nav2 architecture is heading in the right direction, but I don't think this PR is ready to merge yet.

Please address the inline comments first. I also recommend rebasing onto the latest `feature/navigation/nav-core` before requesting review, as the current branch contains merge conflicts (notably around `rover_sensors`). Reviewing after the rebase will reflect the code that will actually be merged.

OShevchuk-UCU (changes requested):
`bridge_gz.yaml` and the `gpu_lidar` block in `rover.urdf.xacro` are shared by both `sim_gz.launch.py` and `sim_gz_nav2.launch.py`, so Nav2-specific `/odom`, `/tf`, and `/scan` bridges and the placeholder `laser_link` also apply to the standard teleop simulation path. On that path, odometry comes from `odometry_controller`, so the `/odom` and `/tf` bridges are unused, but `/scan` is still published because the Gazebo lidar is defined in the shared base URDF—even when Nav2 is not running. This does not break anything today, but it couples two simulation scenarios into one configuration; splitting the bridge and moving the lidar into a sim-only xacro would be cleaner.

OShevchuk-UCU (approved):
LGTM!

BohdanZasymovych (approved):

PR #485, opened 2026-07-02
Refactor/suspension/joystick update

yuriifito:
## Description
PR with many small fixes and joystick-interpreter update  
Fix CAN bus message loss

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- update docs for hotspot (replace deprecated user name with new one)
- fix directory name in host_setup.sh
- add check for can device presence and being up
- refactor joysstick-interpreter node in more oop style to avoid code duplication
- make can udev rule work for all can devices
- Add SocketCAN filters to prevent motor messages from being published to the /from_can_bus topic.

### DIFF
--- a/src/rover_bringup/launch/rover.launch.py
+++ b/src/rover_bringup/launch/rover.launch.py
@@ -1,8 +1,9 @@
 # rover_bringup/launch/rover.launch.py
 import os
+import subprocess
 from ament_index_python.packages import get_package_share_directory
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, RegisterEventHandler, EmitEvent
+from launch.actions import DeclareLaunchArgument, RegisterEventHandler, EmitEvent, OpaqueFunction, LogInfo
 from launch.substitutions import LaunchConfiguration, Command, PathJoinSubstitution
 from launch.events import matches_action
 from launch_ros.actions import LifecycleNode, Node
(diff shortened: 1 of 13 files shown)


src/rover_bringup/launch/rover.launch.py
```diff
@@ -62,11 +82,14 @@ def generate_launch_description():
         parameters=[{
             'interface': LaunchConfiguration('interface'),
             'interval_sec': 0.01,
+            # candump-синтаксис: id:mask (hex, without 0x).
+            # Pass only 0x300-0x3FF (ESP),
+            # all other ids are filtered out.
+            'filters': '300:700,400:700',
```
OShevchuk-UCU: comment/doc say 0x300-0x3FF only, but '400:700' also accepts 0x400-0x4FF (candump OR). Single '300:700' already covers ESP;
drop the second filter or tighten mask to match can_bus.md (0x300-0x30F).

src/rover_teleop/rover_teleop/joystick_interpreter_node.py
```diff
@@ -13,6 +13,74 @@
...
+        self._node = node
+        self._guard = GuardedCall(client)
+        self.name = name
+        self.on = False
+
+    def toggle(self):
+        desired = not self.on
+        req = SetBool.Request()
+        req.data = desired
+        if self._guard.call(req, lambda f: self._on_result(f, desired)):
```
OShevchuk-UCU: self.on set optimistically; _on_result never checks response.success and does not revert on failure/exception (unlike _on_compact_mode_result). Update self.on only when success is True.

src/rover_teleop/rover_teleop/joystick_interpreter_node.py
```diff
@@ -115,93 +151,61 @@ def __init__(self):
...
 
-        cur, pressed = _check_btn(msg.buttons, self._traffic_blue_button, self._prev_traffic_blue_button)
-        if pressed:
-            self._traffic_blue = not self._traffic_blue
-            self._send_traffic()
-        self._prev_traffic_blue_button = cur
+    def _toggle_traffic(self, color: str):
+        attr = f'_traffic_{color}'
+        setattr(self, attr, not getattr(self, attr))
+        self._send_traffic()
```
OShevchuk-UCU: flips local state before send; if _send_traffic's guard rejects the call (busy/not ready), joystick state diverges from hardware.

src/rover_teleop/rover_teleop/joystick_interpreter_node.py
```diff
@@ -330,55 +322,16 @@ def _set_swerve_controller_state(self, activate: bool):
...
-
-    def _on_light_result(self, future, name: str, desired: bool, pending_attr: str):
-        setattr(self, pending_attr, False)
-        try:
-            response = future.result()
-            state = 'ON' if desired else 'OFF'
-            self.get_logger().info(f'{name} {state}: {response.message}')
-        except Exception as exc:
-            self.get_logger().error(f'{name} service call failed: {exc!r}')
+        self._traffic_guard.call(req, self._on_traffic_result)
```
OShevchuk-UCU: return value ignored — warn and/or revert toggles when call() returns False (same pattern as _toggle_motors / _toggle_compact_mode).

OShevchuk-UCU (changes requested):
Good cleanup overall: CAN bring-up checks, generalized udev rule, lighting log fix, and cleaner joystick structure are useful improvements.

I left a few inline comments for issues that are worth fixing before merge:
- receiver filter does not match the documented ESP range;
- traffic-light local state can diverge if the guarded service call is rejected;
- `BoolLight` updates state optimistically and does not handle failed service responses.

The remaining notes are mostly follow-ups: docs alignment, CAN range documentation, validation for `/from_can_bus`, and optional joystick timeout reset handling.

yuriifito:
new fix: return game_controller_node instead of joy_linux. now buttons mapping is right. And also mounted /run/udev in docker, so now SDL (game_controller_node backend) works properly in container and restarts node

OShevchuk-UCU (approved):
LGTM!

PR #490, opened 2026-07-04
feat(arm): add camera_link, servo camera-frame control, keyboard tele…

milka-side:
…op node

## Description

Adds camera_link to the arm URDF, configures MoveIt Servo to use it as the control frame, and implements keyboard-based Cartesian control of the arm relative to the camera frame.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made
- Added `arm_camera_link` to `arm_macro.xacro` as a fixed child of `arm_wrist_2_link`,
  with placeholder box visual geometry and position/orientation offset for the
  physical camera mount.
- Configured MoveIt Servo (`servo.yaml`) to use `arm_camera_link` as the
  command frame for Cartesian control.
- Added `keyboard_servo_node.py` — a ROS 2 node that reads keyboard input and
  publishes `TwistStamped` commands to MoveIt Servo, with:
  - 3 translation axes (forward/backward, left/right, up/down) relative to camera
  - 2 rotation axes (yaw, roll) relative to camera
  - safe-pose routine (`r` key) before starting Servo
  - automatic Servo restart on error/singularity via `/servo_node/status`
- Registered `keyboard_servo_node.py` as an executable in `arm_tasks/setup.py`
- Updated `demo.launch.py` to launch `servo_node` automatically alongside
  `move_group`, so Servo no longer needs to be started with a separate command.
- Core control logic (`ServoController`) is separated from the keyboard input
  layer, so switching to a gamepad (`joy` package) later only requires a new
  input adapter — no changes to `ServoController`.

## Testing

- [ ] All existing tests pass

### How to Test
- ros2 launch arm_moveit_config demo.launch.py
- ros2 run arm_tasks keyboard_servo_node
- Press r to move to safe pose and start Servo
- Use w/s/a/d/q/e for translation, i/k/j/l for rotation

## Additional Notes
Known limitation: pitch axis (u/o keys) is not available due to camera_link having only 5 DOF.


src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+from geometry_msgs.msg import TwistStamped
+from std_srvs.srv import Trigger
+from control_msgs.action import FollowJointTrajectory
+from trajectory_msgs.msg import JointTrajectory, JointTrajectoryPoint
+from builtin_interfaces.msg import Duration
+from std_msgs.msg import Int8
+
+
+# ── Constants ─────────────────────────────────────────────────────────────────
+
```
OShevchuk-UCU, line 48: Hardcoded absolute topic and service names prevent the node from working under ROS namespaces. Consider using relative names instead.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+SERVO_TOPIC = '/servo_node/delta_twist_cmds'
+SERVO_START_SERVICE = '/servo_node/start_servo'
+SERVO_STOP_SERVICE = '/servo_node/stop_servo'
+COMMAND_FRAME = 'arm_camera_link'
+LINEAR_SPEED = 0.1
+ANGULAR_SPEED = 0.3
+PUBLISH_RATE = 50.0
+
+SAFE_POSE = [0.0, 1.2, -1.0, 0.8, 0.5, 0.0]
```
OShevchuk-UCU: These runtime configuration values are hardcoded in the Python node. Consider declaring them as ROS parameters so they can be tuned from launch files/YAML without modifying code.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+        tty.setraw(fd)
+        r, _, _ = select.select([sys.stdin], [], [], timeout)
+        if r:
+            return sys.stdin.read(1)
+        return None
+    finally:
+        termios.tcsetattr(fd, termios.TCSADRAIN, old)
+
+
+def keyboard_loop(controller: ServoController):
```
OShevchuk-UCU: This implementation relies on TTY key auto-repeat for hold-to-move behavior. If true hold-to-move control is intended, consider using a library with explicit key press/release events instead of relying on terminal auto-repeat.

src/arm/arm_description/urdf/arm_macro.xacro
```diff
@@ -143,6 +143,24 @@
...
     <link name="arm_tcp_link"/>
 
+    <!-- Camera mounted on wrist_2 -->
+    <joint name="arm_camera_joint" type="fixed">
+      <parent link="arm_wrist_2_link"/>
+      <child link="arm_camera_link"/>
+      <origin xyz="-0.13286 -0.01058 0.00685" rpy="0 ${-pi/2} 0"/>
+    </joint>
+
+    <link name="arm_camera_link">
```
OShevchuk-UCU, line 171: Consider adding collision geometry for arm_camera_link so MoveIt collision checking accounts for the camera body.

src/arm/arm_moveit_config/config/servo.yaml
```diff
@@ -0,0 +1,53 @@
+moveit_servo:
+  ros__parameters:
+    move_group_name: indomitus_arm
+
+    robot_link_command_frame: arm_camera_link
+
+    ee_frame_name: arm_tcp_link
```
OShevchuk-UCU, line 7: Is it intentional that robot_link_command_frame is arm_camera_link while ee_frame_name remains arm_tcp_link? I want to confirm whether the goal is camera-relative teleoperation (moving the TCP in the camera frame) or using the camera itself as the Servo end-effector frame.

src/arm/arm_moveit_config/launch/demo.launch.py
```diff
@@ -1,7 +1,46 @@
...
+            servo_params,
+        ],
+    )
+
+    demo_launch = generate_demo_launch(moveit_config)
+
+    ld = LaunchDescription()
+    for action in demo_launch.entities:
+        ld.add_action(action)
+    ld.add_action(servo_node)
```
OShevchuk-UCU: Consider delaying servo_node startup until the arm controller is active, or document that Servo expects the demo/control stack to be fully ready before accepting commands.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+        self._stop_client = self.create_client(Trigger, SERVO_STOP_SERVICE)
+        self._traj_client = ActionClient(
+            self,
+            FollowJointTrajectory,
+            '/indomitus_arm_controller/follow_joint_trajectory'
+        )
+        self._timer = self.create_timer(1.0 / PUBLISH_RATE, self._publish)
+
+        self._servo_status = 0
+        self._status_sub = self.create_subscription(
```
OShevchuk-UCU, line 132: /servo_node/status appears to use moveit_msgs/ServoStatus, not std_msgs/Int8.
With Int8 this subscriber will not match the publisher type, so recovery logic will never run.
Please subscribe to moveit_msgs.msg.ServoStatus and use msg.code.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def stop_servo(self):
+        if not self._stop_client.wait_for_service(timeout_sec=2.0):
+            self.get_logger().warn('Servo stop service not available')
+            return
+        self._stop_client.call_async(Trigger.Request())
+
+    def move_to_safe_pose(self):
+        self.stop()
+        self.stop_servo()
```
OShevchuk-UCU: stop_servo() is asynchronous, so this fixed sleep does not guarantee Servo has actually stopped before sending a FollowJointTrajectory goal. Please wait for the stop_servo service response before commanding the safe pose.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+            result = future.result()
+            if result.success:
+                self.get_logger().info('Servo started successfully')
+            else:
+                self.get_logger().warn(f'Servo start failed: {result.message}')
+        except Exception as e:
+            self.get_logger().error(f'Servo start error: {e}')
+
+    def _on_servo_status(self, msg):
+        if msg.data == 2 and self._servo_status != 2:
```
OShevchuk-UCU: Is it intentional to automatically recover only from the singularity status? Should other Servo halt states be handled differently?

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
```diff
@@ -0,0 +1,308 @@
...
+
+    spin_thread = threading.Thread(target=rclpy.spin, args=(controller,), daemon=True)
+    spin_thread.start()
+
+    try:
+        keyboard_loop(controller)
+    except KeyboardInterrupt:
+        pass
+    finally:
+        controller.stop()
```
OShevchuk-UCU, line 596: stop() only publishes zero velocity; it does not disable Servo.
Consider calling stop_servo() before destroying the node so Servo is not left active after exit.

OShevchuk-UCU (changes requested):
Overall, this looks like a solid feature. The separation between `ServoController` and the input layer is a nice design choice and should make it easier to support additional input sources (e.g. a gamepad via the `joy` package) in the future. 

Before merging, I'd like to see the blocking issue with the Servo status message type addressed. The remaining comments are mostly about robustness and maintainability (namespaces, shutdown behavior, keyboard input handling, and documenting the intended MoveIt Servo frame semantics).

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
milka-side: Addressed — using evdev to read raw kernel input events instead of relying on TTY auto-repeat. Key press and release are separate explicit events, so hold-to-move is driven by actual key state, not terminal repeat. termios is only used to suppress terminal echo, not for movement control

src/arm/arm_moveit_config/config/servo.yaml
milka-side: The tool remains the end point for planning, but we linked the command axes to the camera. We chose this approach for intuitive control: since the camera and the gripper share a common base, their orientation now differs only along the Pitch axis. With this setup, the tool's direction of movement in space will always align with what the operator sees on the screen, regardless of how the gripper itself is rotated.
Do you think we should change this strategy?

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
milka-side: If I'm not mistaken, this project is fully on Humble, and in that version moveit_servo publishes /servo_node/status as std_msgs/Int8, not moveit_msgs/ServoStatus — that type change happened in later distros. So I kept the subscriber on Int8.
Confirmed this directly on the running graph:
image
Publisher and subscriber types match (Subscription count: 1), so the recovery logic connects and runs correctly.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
milka-side: SERVO_STATUS_HALT_FOR_SINGULARITY is typically safe to auto-recover — calling start_servo again makes Servo re-attempt the motion and move away from the singular configuration on its own.
Other halts, like HALT_FOR_COLLISION, are deliberately left for the operator instead of being auto-retried — continuing motion there could be unsafe, so the arm needs to be moved clear manually before Servo restarts.

src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
OShevchuk-UCU: Thanks for confirming. That answers my concern. Resolving.

src/arm/arm_moveit_config/config/servo.yaml
OShevchuk-UCU: No, I think the current strategy makes sense, as it provides more intuitive teleportation while keeping the TCP as the planning end-effector.

OShevchuk-UCU (approved):
Thanks for addressing the review comments. The issues I raised have been resolved, and the implementation looks good to me.

One note for future PRs: please fix any CI failures before requesting a review. Otherwise, reviewers may end up reviewing the same PR multiple times, creating unnecessary duplicate work.

Approving from the code review perspective. The remaining CI issue should be resolved before merging.

Martik-k (approved):

PR #491, opened 2026-07-05
Refactor/suspension/tf and sim

yuriifito:
## Description
Small fixes with TF. Add separate hardware component for differential bar and controller for it in simulation

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- remove publishing to odom tf from wheels odometry (robot_localization should do that instead)
- Fix: replace unidirectional rocker position-mirroring with bidirectional torque-based differential coupling
- refactor diff bar node into controller (as it works with joints)


OShevchuk-UCU (changes requested):
Thanks for the refactor. The diff-bar move into a ros2_control controller looks like the right direction.

However, I think this PR should not be merged standalone yet. It removes the odom→base_link TF publisher from the odometry controller, but does not add robot_localization or any other replacement TF publisher. After this merge, the rover/sim stack appears to have no owner for odom→base_link.

Also, /odom is renamed to /odometry/wheels without a migration path or documentation update, which may break existing consumers.

One more major issue: RockerDiffController relies on hardcoded state_interfaces_[0..3] and command_interfaces_[0..1]. Please validate/map interfaces by name during activation to avoid silently applying torque to the wrong joints if interface ordering changes.

yuriifito:
What's done:
- robot_localization package added, EKF node configured to merge /odometry/wheels (currently the only source).
- launch file created (rover_localization/launch/ekf.launch.py), parameterized with use_sim_time
- added into both rover_sim/launch/gz_sim.launch.py and rover_bringup/launch/rover.launch.py.
- added validation of state and command interfaces by name in RockerDiffController
- added covariance for wheels odometry

yuriifito:
I want to refactor launch files structure in next PR, but there are questions:
1. where should I store launch file for `twist_mux`? keep it in bringup or make separate package for it for instance rover_control?
2. where would it be better to move can launch file? to separate rover_can package? because bringup package doesn't seem like the right place for it

OShevchuk-UCU (changes requested):
Thanks for addressing the main review points. The ownership of `odom → base_link`, interface mapping in `RockerDiffController`, and odometry covariance now look correct.

I found only a few remaining items before approval:

* `rover_localization/package.xml` should declare `robot_localization` as an `exec_depend`.
* `rover_bringup` and `rover_sim` should declare `rover_localization` as an `exec_depend`, since their launch files now use it.
* The documentation should be updated to remove `sim_diff_bar_node`, document the EKF and `/odometry/wheels`, and mention the `/odom` topic migration.

I also noticed that the wheel geometry parameters differ between simulation and hardware:

* `track_width`: `0.689` in simulation vs `0.682` on hardware
* `wheel_radius`: `0.15` in simulation vs `0.157` on hardware

Please verify whether this difference is intentional. It does not necessarily need to block this PR, but it should either be aligned or tracked in a follow-up issue.

Regarding the launch-file structure:

* I would keep the `twist_mux` launch file in `rover_bringup` for now. It is system-level integration and does not belong to the controller implementation itself. A separate package would make sense later if multiple mux configurations or autonomy command sources are introduced.
* Moving the CAN launch file into a dedicated `rover_can` or `rover_drivers` package sounds reasonable, especially if the CAN stack will include the gateway, device configuration, diagnostics, and ID mappings. This can be done in the next PR; keeping it in `rover_bringup` temporarily is acceptable.

After the dependency and documentation updates, I think this PR can be approved with the geometry and launch restructuring handled as follow-up work.

yuriifito:
having different geometry parameter for sim rover and for real one is intentional, because in reality we have wheels of different radius for now and suspension geometry also differs, and we should fix it in next suspension iteration

OShevchuk-UCU (approved):
Thanks for the updates and clarifications. LGTM, approving.

PR #493, opened 2026-07-08
Feature/navigation/aruco

yagodanr:
…amera from pc

## Type of Change

- [ X ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- Added Aruco marker detection launch file
- Added launch version with laptop camera node 
- Added documentation
- Added parameter file
- Given permissions in `docker-compose` file for the camera, but commented out

## Testing

- [ X ] I have tested this locally
- [ ] All existing tests pass

### How to Test

1. Uncomment video permission lines in docker-compose
2. `docker compose up --build`
3.  `ros2 launch rover_localization aruco_debug.launch,py`


yagodanr:
I have a question. Should debug launch file with laptop_camera node be included in repo and where? I also had to change docker-compose file for video permissions. Maybe it shouldn't be included in repo, but otherwise someone else would have to spend their time trying to test with laptop/USB camera

OShevchuk-UCU:
> I have a question. Should debug launch file with laptop_camera node be included in repo and where? I also had to change docker-compose file for video permissions. Maybe it shouldn't be included in repo, but otherwise someone else would have to spend their time trying to test with laptop/USB camera

I think the debug launch should stay in the repository because it provides a reproducible bench-test workflow and saves time for other developers. I'd just keep it clearly marked as a debug/development launch. The commented Docker camera permissions are also fine in the example compose file—they shouldn't be enabled by default.

OShevchuk-UCU (changes requested):
Overall, this looks good. I left a few comments, mostly around making the main launch more reusable while keeping the laptop/USB camera workflow as a dedicated debug path.

---

Make the main ArUco launch camera-agnostic

The main `aruco.launch.py` is intended to work with an already running camera node, but it is currently coupled to a specific camera topic (`/camera/image_raw`) through the default parameter file.

Consider exposing the camera topic (or base topic) as a launch argument/parameter override and keeping the laptop USB camera configuration only in `aruco_debug.launch.py`. This will make the main launch reusable with the rover camera pipeline and other camera drivers without modifying the package configuration.

---

Avoid hardcoded USB camera device

`/dev/video0` works for many systems, but camera enumeration may differ across machines.

Consider exposing the device path as a launch argument so developers can use a different camera without editing the package configuration.

---

Add a short verification checklist

The documentation explains how to start the nodes, but it would also be useful to describe how to verify that the setup works correctly.

For example:
- verify `/camera/image_raw`
- verify `/camera/camera_info`
- verify `/aruco_detections`
- verify TF output (if enabled)

PR #496, opened 2026-07-11
Feature/suspension/current voltage viz

yuriifito:
## Description
Add tools for node measurements and analysis

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- Add new node that subscribes to `/from_can_bus` topic and takes all frames with ids 0x302 or 0x303
- Add plotjuggler to visualize current and voltage data

## Testing

- [x] All existing tests pass
- [ ] I have added new tests


OShevchuk-UCU (changes requested):
Overall, LGTM. I only have a few follow-up suggestions:

- The PlotJuggler layout currently visualizes only `sensor_rover`, while the node publishes data for both configured sensors. It would be useful to include `sensor_arm` as well.
- Consider validating `sensor_can_ids` for duplicate values during startup, since duplicate IDs would silently overwrite entries in `_sensors_by_id`.
- Please verify that the required PlotJuggler ROS 2 plugin/package is declared, so the bundled layout works on a clean Humble installation.

yuriifito:
We don't have arm power sensor yet, it will be added in a week. So we cannot now add arm sensor plot till there is no data being published on that topic

PR #497, opened 2026-07-11
Feature/navigation/slam integration

MunenNoKage:
## Description
This PR replaces the old static map and AMCL localization with a live SLAM mapping setup. It also adds a real 3D model for the RPLIDAR S2 sensor, switches to a proper odometry controller, and tunes navigation behaviors.

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Configuration change
- [x] Code refactoring

### SLAM & Maps
* **Removed AMCL:** Deleted the old static test map files and removed AMCL from the navigation launch.
* **Added Live SLAM:** Added `slam_toolbox` config and launch files to build maps in real-time.
* **New Planner:** Swapped the global path planner from `NavfnPlanner` to `SmacPlanner2D`.

### Lidar & URDF
* **3D Mesh:** Added the official `rplidar_s2.stl` 3D model mesh.
* **Sensor Specs:** Updated the lidar range to track from 0.05m up to 18m (instead of 0.12m to 10m).

### Odometry & Code
* **Instant Odom:** Fixed `odometry_controller.cpp` to publish the starting position immediately so the system doesn't lag at startup.
* **Cleaned Bridges:** Removed the old Gazebo ground-truth odom plugin and topics. The `odometry_controller` is now spawned directly.

### Navigation & BT Tuning
* **Controller Updates:** Tuned the MPPI controller parameters (costs, alignment).
* **Behavior Tree Fix:** Fixed issues with the "Navigate Through Poses" behavior tree.

### RViz Workspace
* **Clean UI:** Removed unneeded panels (Time, Selection, Views). Only the Displays list and Nav2 panel are left.

### How to Test
Run these commands in separate terminals to test the live SLAM and navigation stack:

1. Launch the simulation:
```bash
ros2 launch rover_sim sim_gz_nav2.launch.py
```
2. Launch SLAM:
```bash
ros2 launch rover_navigation slam.launch.py use_sim_time:=true
```

3. Launch navigation:
```bash
ros2 launch rover_navigation navigation.launch.py
```

4. Launch RViz:
```bash
ros2 launch rover_viz rviz.launch.py use_sim:=true use_nav:=true
```

### Additional Notes
The map will now generate dynamically as the rover drives around


OShevchuk-UCU (approved):
Overall, this looks like a solid improvement to the navigation stack. The transition to SLAM Toolbox, the odometry changes, and the Nav2 tuning all make sense. I only have a couple of minor comments before merging.

### 1. Behavior Server parameters

I think `lim_linear_accilation` / `lim_linear_deceleration` are not valid Nav2 Humble `behavior_server` parameters, and `accilation` also looks like a typo. Could you verify these against the Nav2 Humble parameter names? Otherwise they may be silently ignored and the intended tuning will not take effect.

### 2. Documentation

Please update the project documentation to reflect the new SLAM-based workflow, not only the PR description, so future users follow the correct launch sequence and setup.

PR #498, opened 2026-07-11
Refactor/suspension/launch restructure

yuriifito:
## Description
Restructure launch files

## Type of Change

- [x] Bug fix
- [x] Code refactoring

## Changes Made

- Refactor rover.launch.py by encapsulating logic into smaller, composable files
- Fix bug in can launch py (after fail to launch can because interface didin't exist, several ros2_socketcan nodes were still alive even after a long time)
- move launch files for container, lighting to rover_peripherals package
- move joy.launch.py to rover_teleop node (it logically belongs there)

## Testing

- [x] All existing tests pass


OShevchuk-UCU (approved):
Overall, I think this refactoring is moving the project in the right direction. Splitting the launch files by subsystem and moving them closer to the packages that own the corresponding nodes makes the launch hierarchy much easier to understand and maintain.

I only have a few architectural suggestions for follow-up:

- Ensure that `package.xml` dependencies stay in sync with the launch include graph. If a package includes launch files from another package, it should declare the corresponding runtime dependency.
- Consider extracting the remaining inline `ros2_control` stack into a dedicated `control.launch.py` so that `rover.launch.py` becomes a thin composition layer.
- The `include_launch()` helper is now duplicated across multiple launch files. It may be worth moving it to a shared utility module if this refactoring continues.
- Don't forget to update Docker entrypoints and project documentation to reflect the new package ownership (e.g. `joy.launch.py` now lives in `rover_teleop`).

Overall, I like the direction of this refactoring and think it will make the launch architecture much easier to extend as the project grows.

yuriifito:
## Summary

**Node fixes**
- `rover_power_node.py`: guard `rclpy.shutdown()` with `rclpy.ok()` (fixes `RCLError` on Ctrl+C); validate `sensor_can_ids` for duplicates

**Xacro unification**
- Merged real/sim `ros2_control` and top-level xacro files into `rover.xacro` + `rover.ros2_control.xacro`, branched via `xacro:if`/`xacro:unless` on new `use_sim` arg

**Launch refactor**
- Added `rover_description/launch/robot_state_publisher.launch.py` and `rover_bringup/launch/control.launch.py`, shared between real and sim (`control.launch.py` skips `ros2_control_node` when `use_sim:=true`, since Gazebo's plugin owns it there)
- Simplified `rover.launch.py` / `sim_gz.launch.py` to compose the shared launch files instead of duplicating logic
- Add `twist_mux` launch file
- Add `robot_state_publisher.launch.py`
- Move launch files with rover_peripherals nodes to its package

**Docs**
- Updated nodes/packages overview to match current structure
- Added `tf_ownership.md`

OShevchuk-UCU (approved):
LGTM!

PR #501, opened 2026-07-16
Feature/arm/gazebo moveit camera sim

milka-side:
## Description

Adds a full Gazebo simulation for the arm, integrated with the existing
MoveIt/ros2_control stack: the arm spawns in Gazebo with gz_ros2_control
driving the same `indomitus_arm_controller` used elsewhere, move_group and
moveit_servo run on Gazebo's sim time, the wrist camera is simulated and
bridged to ROS, and RViz can visualize the running sim. Keyboard servo
teleop now works against the sim end-to-end.

Because the CAD meshes (~500k triangles) made the simulated camera
unaffordable to render without GPU acceleration (real-time factor dropped
to ~0.5%), the sim uses a new primitive-geometry variant of the arm
(boxes/cylinders, identical joints, masses and inertias) — RTF is ~100%
with it. The mesh-based description remains the default for RViz and real
hardware.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- `arm_sim`: Gazebo world, `gz_bridge.yaml` (clock + camera topics),
  `ros2_controllers.yaml`, and `arm_gazebo.launch.py` (Gazebo +
  ros2_control + move_group + servo, ordered startup: spawn → controller
  spawner → move_group/servo, all on sim time)
- `arm_gazebo_rviz.launch.py`: one-command Gazebo + RViz (reuses the
  moveit_rviz setup on sim time; replaces running demo.launch.py next to
  Gazebo, which duplicated controller_manager and /joint_states)
- `arm_description`: primitive-geometry variant (`arm_macro_simple.xacro`,
  `arm_standalone_simple.urdf.xacro`) used by the Gazebo launch; `camera:=false`
  arg to drop the camera sensor; camera orientation fix and Gazebo visual colors
- `keyboard_servo_node`: wait for the safe-pose trajectory result instead of
  an 8 s wall-clock timeout (always fired at low RTF), report rejected goals,
  raise default speeds to 0.25 m/s / 0.75 rad/s

## Testing

- [ ] All existing tests pass

### How to Test

1. `colcon build --symlink-install && source install/setup.bash`
2. `ros2 launch arm_sim arm_gazebo_rviz.launch.py` — Gazebo and RViz open,
   controllers activate, "You can start planning now!" appears
3. `ros2 run arm_tasks keyboard_servo_node --ros-args -p use_sim_time:=true`, press `r` — arm moves to the
   safe pose in ~3 s, then drive with `w/a/s/d/q/e` (translation) and
   `i/k/u/o/j/l` (rotation); motion is visible in both Gazebo and RViz
4. Camera: `ros2 topic hz /camera/image_raw` (~30 Hz)
5. Sim speed: RTF in the Gazebo status bar should be ~100%

## Additional Notes

- The sim intentionally uses primitive geometry; visual fidelity is
  traded for real-time factor. Mesh-based description is unchanged for
  all non-sim use.
- MoveIt still plans against the mesh model; frames are identical so
  planning/servoing are unaffected.
- Known quirk: `camera:=false` only exists in the simple macro
  (`arm_macro.xacro` ignores it).


OShevchuk-UCU (changes requested):
Thanks for adding the Gazebo, MoveIt, Servo, and camera simulation stack. Overall, the architecture looks good: Gazebo owns `controller_manager` through `gz_ros2_control`, the controller startup is event-based, and the RViz wrapper avoids starting a second mock ros2_control stack.

Before merging, could you please verify and address the following points:

1. **Simulation time for keyboard teleop**

   `keyboard_servo_node` is started separately using:

   ```bash
   ros2 run arm_tasks keyboard_servo_node
   ```

   but its published Twist messages are stamped using the node clock, while MoveIt Servo runs with `use_sim_time:=true`.

   Please either:

   * add a simulation launch file that starts the keyboard node with `use_sim_time:=true`, or
   * update the documentation to use:

   ```bash
   ros2 run arm_tasks keyboard_servo_node --ros-args -p use_sim_time:=true
   ```

2. **Safe-pose action result handling**

   The new implementation waits for the trajectory result, but it does not inspect the returned action status or controller error code. An aborted or canceled trajectory can therefore still be reported as:

   ```text
   Safe pose reached!
   ```

   Please verify the result status before returning success, and consider adding a configurable wall-clock timeout so the worker cannot wait indefinitely if the controller stops responding.

3. **Default speed change also affects real hardware**

   The default keyboard Servo speeds were increased from `0.1 / 0.3` to `0.25 / 0.75` in the shared node.

   Since this PR is primarily simulation-related and is marked as non-breaking, please keep the existing hardware defaults and override the higher values only in the simulation launch/configuration.

4. **`camera:=false` behavior**

   The camera sensor is conditionally removed from the Xacro, but the bridge still starts with `/camera/image_raw` and `/camera/camera_info` entries.

   Please verify the runtime behavior with:

   ```bash
   ros2 launch arm_sim arm_gazebo_rviz.launch.py camera:=false
   ```

   It should start without repeated bridge errors or misleading camera-related failures. If it produces warnings continuously, please conditionally use a clock-only bridge configuration when the camera is disabled.

5. **Launch failure propagation**

   Controllers currently start whenever the spawn process exits, and MoveIt/Servo start whenever the controller spawner exits, regardless of the return code.

   Please verify failure scenarios and, if necessary, gate the next startup stage on a successful exit or shut down the launch when spawning/controller activation fails.

6. **Clean-install dependencies**

   Please verify the package from a clean environment and ensure `arm_sim/package.xml` explicitly declares the runtime controller plugins and Python dependencies used directly by this package, particularly:

   * `joint_trajectory_controller`
   * `joint_state_broadcaster`
   * `moveit_configs_utils`
   * Python YAML support

7. **Camera frame convention**

   Please verify that `/camera/camera_info` uses the expected camera frame and that the frame follows the ROS optical convention. If this camera will later be used for perception rather than only visualization, it would be better to add a dedicated `arm_camera_optical_frame`.

The intentional primitive-model versus mesh-model difference looks acceptable for performance, provided it is documented that joint transforms are equivalent but collision geometry is not necessarily identical.

Once these points are verified and the necessary fixes are applied, I do not see any additional blockers from my side and the PR can be merged.

Martik-k (approved):

OShevchuk-UCU (changes requested):
Thanks for the follow-up — the main review items look addressed. Two remaining non-blocking points:

### 1. Safe-pose failure still starts Servo

In `keyboard_servo_node.py`, `_handle_safe_pose` always calls `start_servo()` after `move_to_safe_pose()`, regardless of the return value:

```python
self._controller.move_to_safe_pose()
print('Starting servo...')
self._controller.start_servo()
```

`move_to_safe_pose()` now correctly returns `False` on reject / abort / controller error / timeout, but Servo is still started afterward. That means the operator can resume Cartesian teleop from a pose that never reached the intended safe configuration (e.g. after an aborted trajectory).

Suggested change:

```python
if self._controller.move_to_safe_pose():
    print('Starting servo...')
    self._controller.start_servo()
else:
    print('Safe pose failed — Servo not started.')
```

### 2. “Simple” Gazebo model docs no longer match the active geometry

`arm_standalone_simple.urdf.xacro` still describes this variant as primitive boxes/cylinders for cheap rendering. After `fix(sim): adjust spawn height for arm_mount mesh origin`, `arm_macro_simple.xacro` has **mesh** visual/collision active for most links (mount, base, shoulder, forearm, wrists), with the primitive blocks commented out. Only the end-effector path still looks primarily primitive.

So either:

- update the standalone / macro comments to say this file is currently a mesh-based Gazebo description (with optional primitives left commented for perf experiments), or
- switch the primitives back on if that was still the intended default for RTF.

As written, the filename + comments suggest a light collision/visual model, but the spawned Gazebo robot is largely mesh again — including the collision mismatch / performance tradeoff discussion from earlier.

Neither of these blocks merge from my side if you prefer to track them separately.

OShevchuk-UCU (approved):
LGTM!

PR #511, opened 2026-07-30
feat(arm): add gamepad teleop via ros2 joy game_controller_node

milka-side:
## Description

Adds gamepad teleop for the arm as an alternative to keyboard control,
and hardens it for real hardware use. `ServoController` (the shared
node powering Cartesian velocity commands and the safe-pose action) is
unchanged — a new `GamepadInputLoop` class subscribes to `/joy`
(published by `ros2 joy`'s `joy_node`) instead of reading raw keyboard
events, exactly as the module's docstring already said this
integration should work.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made

- `GamepadInputLoop`: right stick → forward/back + left/right, left stick → yaw/pitch, L2/R2 → roll, holding **Y** shifts L2/R2 to
  up/down instead, **A** → move to safe pose + start servo, **X** → exit
- Anti-jump guard on the L2/R2 shift: if the triggers are already off-center when Y is pressed, the shifted axis stays at 0 until they
  pass back through center once — switching modes never causes a sudden velocity jump
- Live feedback printed on control-state transitions (which stick/trigger is active + resulting vx..wz), mirroring `KeyboardInputLoop`'s key-down print
- **`/joy` watchdog**: a timer independent of message arrival stops the arm if no `Joy` message has been received for 0.2s, since `ServoController` otherwise keeps republishing the last commanded twist forever if the controller disconnects
- **Safe-pose input lock**: stick/trigger input is ignored while a safe-pose move is in progress, so a held stick can't resume Cartesian motion the instant Servo restarts
- One-time arm gate: teleop stays locked out until the first safe-pose move succeeds (Servo isn't started before that anyway); a later `/joy` dropout does **not** re-lock it or force another trip to the safe pose — control resumes from wherever the arm already is, gated only by requiring one genuinely centered reading before trusting axes again (protects against known BLE reconnect glitches where stale/garbage values are reported before the first real update)
- Button presses are only recognized as edges against a real prior reading — the first message after startup or a reconnect is used solely to establish a baseline, since a gamepad can report a stale "pressed" button on that first message (was causing spurious exits)
- Uses `ros2 joy`'s `joy_node` rather than `game_controller_node`: `game_controller_node` goes through SDL, which has no built-in axis mapping for this controller's Bluetooth identity and falls back to an auto-generated one that's broken (right stick unmapped, triggers pointed at the wrong axes). `joy_node` reads the kernel's already-calibrated joystick device directly, sidestepping that whole class of bug — no hardcoded mapping needed
- Trigger axes (L2/R2) rest at +1.0 and go to -1.0 at full press under `joy_node` (opposite convention from the sticks); `_trigger_amount()` normalizes that back to "0 at rest" so deadzone, the settle-guard, and the Y-shift logic don't need to special-case them
- Deadzone raised from 0.15 to 0.2
- `KeyboardInputLoop`: direction keys are now ignored entirely until Servo has actually started — pressing them earlier couldn't move the arm anyway, and `_handle_safe_pose` discards any accumulated key state regardless
- New `gamepad_servo_node` console entry point; `joy` added to `arm_tasks/package.xml`
- Reuses `ServoController` unchanged, so all existing behavior carries over automatically: safe-pose result verification + timeout, hardware vs. sim speed defaults via the existing `arm_sim/config/keyboard_servo_sim.yaml` params file

## Testing

- [ ] All existing tests pass

### How to Test

1. `colcon build --symlink-install && source install/setup.bash`
2. `ros2 run joy joy_node` (confirm it logs `Opened joystick: ...`)
3. `ros2 run arm_tasks gamepad_servo_node` (or add `--ros-args --params-file $(ros2 pkg prefix arm_sim)/share/arm_sim/config/keyboard_servo_sim.yaml` for Gazebo sim time + faster speeds)
4. Move each stick, hold Y while moving L2/R2, press A (safe pose + servo start) and X (exit); watch the printed feedback line match what you're doing
5. Confirm sticks/triggers are ignored until A is pressed, and again during a safe-pose move
6. Disconnect/reconnect the controller mid-session and confirm the arm stops on disconnect and only resumes once a clean, centered reading comes through (all sticks at 0, both triggers released) — no phantom motion
7. To find/confirm axis and button indices for a specific controller: `ros2 topic echo /joy` while moving each control one at a time

### DIFF
--- a/src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
+++ b/src/arm/arm_tasks/arm_tasks/keyboard_servo_node.py
@@ -41,8 +41,8 @@ Usage:
         ros2 run arm_tasks gamepad_servo_node
 
     ``gamepad_joy_driver`` is a thin wrapper around ``ros2 run joy
-    joy_node`` — see ``main_gamepad_joy_driver`` below for why this is
-    used instead of ``game_controller_node``.
+    joy_node``, kept as its own entry point so this launch command stays
+    stable regardless of what runs underneath it.
 """
 
 import os
@@ -482,6 +482,9 @@ class KeyboardInputLoop:
         self._exit_event = threading.Event()
         self._device = None
         self._read_thread = None
+        # Direction keys are ignored until Servo is running (see
+        # _read_loop) — pressing them earlier can't move the arm anyway.
+        self._servo_started = False
 
     def _open_device(self) -> bool:
         """Open the evdev keyboard device at ``self._device_path``.
@@ -554,6 +557,7 @@ class KeyboardInputLoop:
         if self._controller.move_to_safe_pose():
             print('Starting servo...')
             self._controller.start_servo()
+            self._servo_started = True
         else:
             print('Safe pose failed — Servo not started.')
 
@@ -588,6 +592,14 @@ class KeyboardInputLoop:
                     continue
 
                 if code not in self._DIRECTIONS:
+                    continue
+
+                # Direction keys are ignored entirely until Servo has
+                # started — _handle_safe_pose clears _pressed anyway, so
+                # tracking presses before that point would just be
+                # discarded, and set_velocity()'d twists Servo isn't
+                # listening to yet would have nothing to show for it.
+                if not self._servo_started:
                     continue
 
                 if value == self._KEYSTATE_DOWN:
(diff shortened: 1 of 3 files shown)


OShevchuk-UCU (changes requested):
Overall, this is a well-structured change. I like that `ServoController` remains unchanged and `GamepadInputLoop` is implemented as an alternative input source, with the shared `_run_teleop()` reducing code duplication. The safe-pose lock and defensive handling of short `Joy` messages are also nice touches.

I do have a couple of concerns before merging:

- **Safety on `/joy` loss.** Since `ServoController` continuously republishes the last commanded twist, losing the joystick (e.g. controller disconnect or `joy` node failure) could leave the arm moving indefinitely because Servo's command timeout never expires while commands continue to be published. I think adding a watchdog that stops the controller if no `Joy` message is received for a short period would make this much safer and more robust for real hardware.

- **Input during the safe-pose sequence.** While the safe-pose action is running, `Joy` callbacks continue updating commanded velocities. If the operator is holding the sticks, Cartesian motion may resume immediately after `start_servo()`. It would be safer to temporarily ignore joystick input (or force zero velocities) until the safe-pose sequence has fully completed.

The remaining comments I have are mostly smaller improvements (e.g. configurable controller mappings and a few edge cases around trigger mode switching), but I think the two points above are worth addressing before merge. Other than that, the implementation looks clean and the overall approach makes sense.

OShevchuk-UCU (approved):
Looks good. Both merge blockers are addressed.

PR #513, opened 2026-08-03
Feature/suspension/motors state poll

yuriifito:
## Description

Makes motor and CAN-bus faults observable and diagnosable after the fact.

Previously a motor fault was visible only as a `fault_code` byte on a sampled
status topic. That has two problems: the vendor encodings are mutually
incompatible (Damiao packs state *and* fault into one nibble, Steadywin uses a
bitmask plus a separate mode byte), so every consumer had to know which motor it
was talking to.

This PR introduces a vendor-neutral fault taxonomy, edge-triggered fault events
carrying a freeze frame of motor state at the instant of the transition, and an
on-disk log so a fault that happened in the field can be explained afterwards.
The CAN bus itself is treated as a faultable component and flows through the
same pipeline, so bus-off and error-passive conditions are reported once against
the interface rather than eight times over as phantom motor failures.

Along the way the SocketCAN handling and the fault bookkeeping are lifted out of
`RoverHardwareInterface` into `CanBus` and `FaultDetector`, which is what makes
the above testable without hardware.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update
- [x] Configuration change

## Changes Made

**Vendor-neutral fault taxonomy** — new `motor_fault.hpp`

- `rover_fault::Fault` enum: 15 normalised values, including two transport-level ones (`CAN_ERROR_PASSIVE`, `CAN_BUS_OFF`) reported against the interface, not against a motor.
- `rover_fault::Recovery` policy enum + `recovery_for()` — recovery is a property of the fault class, not of the vendor that reported it. Physical faults (encoder, hardware) latch and are never auto-retried.
- Each vendor header now owns a `translateFault()`; nothing above that layer branches on vendor.
- Damiao: ERR nibble mapped; **`0x0`/`0x1` are operating states, not faults** — `0x1` (enabled) was previously being read as a fault. Reserved codes `0x2`–`0x7` map to `UNKNOWN` rather than `NONE`, so a firmware change cannot silently look healthy.
- Steadywin: bitmask + mode byte, most-severe-wins. **Mode 0 (OFF) is now detected** — the motor sets no fault bit for that condition, so a check on `fault_code` alone read a dead motor as healthy.

**Edge-triggered fault events** — new `indomitus_interfaces/msg/FaultEvent.msg`, published on `/fault_events`

- Five transitions: `FAULT_ENTER`, `FAULT_CHANGE`, `FAULT_CLEAR`, `SIGNAL_LOST`, `SIGNAL_OK`. One message per transition, never periodic state.
- Carries a **freeze frame** (position, velocity, torque, temperature, voltage, current, mode, enabled) captured at the moment of the transition — the values that explain *why*, before they recover.
- Carries the untranslated `raw_code` alongside the normalised fault, for post-hoc analysis.

**`CanBus`** — new `can_bus.hpp` / `can_bus.cpp` (extracted from `RoverHardwareInterface`)

- Owns socket lifecycle, a dedicated blocking RX thread, and CAN error-frame decoding.
- Exposes bus state (`OK` / `ERROR_WARNING` / `ERROR_PASSIVE` / `BUS_OFF`), TX/RX error counters, and a dropped-frame count for a full TX queue.
- Knows nothing about motors or protocols. Frame routing stays in the hardware interface, since that is protocol knowledge.

**`FaultDetector`** — new `fault_detector.hpp` / `fault_detector.cpp` (extracted)

- Edge-triggers for all 8 motors plus the bus; queues events (capped at 256) for a publisher to drain from another thread.
- No I/O and no ROS publishing on the detection path — `detect_*()` is called once per control cycle from the RT `read()` thread.

**`RoverHardwareInterface`**

- CAN and fault logic removed; `NUM_WHEELS` moved to a new `constants.hpp`.
- `motors_enabled_` is now `std::atomic<bool>` — it is written by service callbacks and lifecycle transitions and read from the control thread and three timer callbacks, so the plain `bool` was a data race.
- New `activated_` flag: `read()` is called by `controller_manager` while the component is merely *configured*, before `on_activate()` has opened the socket, so "socket not open" was reported as a fault when it was not one.
- Fault publishing deferred to a 20 Hz timer on `hw_node_` rather than done inline — `rclcpp` publishing allocates, and a stall in `read()` longer than the Damiao `TIMEOUT` register would provoke the very comm-loss fault being reported.

**New `rover_diagnostics` package**

- `fault_logger_node` — subscribes `/fault_events`, appends each transition to a rotating JSONL file with its freeze frame. Lives outside the hardware interface deliberately: writing to disk from the update loop risks the same stall described above.
- `event_log.py` — append-only JSONL, **fsync per event** (the failure being explained is often a brownout, and a buffered write loses exactly the last few events) and size-capped rotation (a full disk is itself a rover failure mode). NaN/Inf sanitised to `null`, since bare `NaN` is not valid JSON.
- `fault_logger.launch.py` with a `log_dir` argument (default `~/.ros/rover_faults`).

## Testing

- [x] All existing tests pass
- [x] I have added new tests (if applicable)


OShevchuk-UCU (changes requested):
## Review summary

The overall direction of this PR is good: introducing a vendor-neutral fault taxonomy, edge-triggered fault events, CAN-bus diagnostics, freeze-frame logging, and separating CAN/fault handling from `RoverHardwareInterface` significantly improves the architecture and testability.

However, I do not think the PR is ready to merge yet because several issues affect shutdown safety, real-time behavior, and the reliability of the fault reporting itself:

* `CanBus::stop_rx()` may block indefinitely because it calls `join()` while the RX thread can still be blocked in `read()`. The socket must be used to unblock the receive thread before waiting for it to exit, or the RX loop should use `poll()` with an `eventfd`/timeout.
* Motor feedback validity appears to remain set after the first received frame. Without a feedback timeout that invalidates stale state, a motor that stops transmitting may never generate `SIGNAL_LOST`.
* The first received feedback may incorrectly generate `SIGNAL_OK`, even though no preceding signal-loss transition occurred.
* The SocketCAN descriptor is blocking despite the comment that sending is non-blocking. A blocked `write()` from the control path could stall the `ros2_control` loop and trigger additional communication faults. The socket should use `O_NONBLOCK`, with explicit handling of `EAGAIN` and `ENOBUFS`.
* `fd_` is accessed from RX, TX, and lifecycle threads without a clear synchronization/lifetime strategy, creating a potential race during deactivate or shutdown.
* The fault-detection path still performs operations that are undesirable in the real-time `read()` loop, including mutex locking, `std::vector` growth/erase, message/string construction, and `std::function` use. The healthy no-transition path should ideally perform no heap allocation, and a fixed-capacity queue would be safer.
* CAN `ERROR_WARNING`/`ERROR_PASSIVE` may remain sticky because recovery through `CAN_ERR_CRTL_ACTIVE` does not appear to restore the state to `OK`.
* `TRANSIENT_LOCAL` history can replay old events when the logger restarts, causing duplicate entries in the forensic JSONL log unless an event/session ID or deduplication mechanism is added.
* Unknown Steadywin fault bits currently appear capable of being interpreted as healthy. Reserved bits should preferably map to `UNKNOWN`, consistently with the Damiao implementation.
* The fault logger should be included in the production bringup; otherwise the new persistent logging functionality depends on someone launching it manually.
* The unrelated acceleration, deceleration, steering-range, Docker device, and teleoperation changes should be separated from this diagnostics PR.

Before merge, I recommend fixing at least the RX-thread shutdown ordering, feedback timeout and signal transitions, non-blocking CAN transmission, descriptor synchronization, CAN-state recovery, and allocations/locking on the real-time detection path.

It may also be worth adding tests covering:

* stopping the RX thread with no incoming CAN traffic;
* feedback timeout followed by `SIGNAL_LOST` and `SIGNAL_OK`;
* first-observation behavior;
* CAN warning → passive → active/recovered transitions;
* full TX queue and interface-down behavior;
* repeated activate/deactivate cycles;
* queue overflow and concurrent draining;
* Steadywin reserved bits and Damiao codes including `0xF`;
* logger restart/replay, disk-full, and rotation scenarios.

OShevchuk-UCU (approved):
## Review follow-up

The previous blockers appear to be addressed: `stop_rx()` now wakes cleanly via `eventfd`/`poll`, feedback freshness and `SIGNAL_LOST` / `SIGNAL_OK` handling are improved, TX is non-blocking, `fd_` access is synchronized, CAN recovery states are handled, vendor fault translation is safer, logger replay is deduplicated, and useful unit tests were added.

The remaining items are relatively small:

1. **Add a defensive DLC check in `CanBus::send()`.**
   `memcpy(frame.data, data, dlc)` currently assumes `dlc <= 8`. Existing callers appear valid, but rejecting `dlc > CAN_MAX_DLEN` would be a small and worthwhile memory-safety hardening.

2. **RT path is much better, but some follow-up cleanup remains.**
   Fault detection avoids work on the healthy/no-transition path. On transitions, however, the queue still uses a mutex, strings are copied/assigned, and `drain_events()` allocates/copies into a `std::vector`. This is acceptable for rare fault edges and is not a merge blocker, but a lock-free/preallocated approach could be considered later.

3. **Additional tests would still be valuable as follow-ups:**

   * actual `CanBus` error-state recovery (`ERROR_WARNING → ERROR_PASSIVE → ACTIVE`, `CAN_ERR_RESTARTED`);
   * TX backpressure / `ENETDOWN`;
   * full hardware lifecycle reactivation (`configure → activate → deactivate → activate`);
   * logger write failures and rotation edge cases such as `backup_count == 0`.

4. **Optional longer-term API hardening:** an `event_id` / `boot_id` in `FaultEvent` would make forensic correlation and replay deduplication more robust across sessions.

### Merge stance

**Looks good to merge after the small DLC bounds check.** The RT queue improvements, additional CAN/lifecycle/logger tests, and event identity hardening can reasonably remain follow-up work.

PR #514, opened 2026-08-06
Feature/suspension/radius mode

yuriifito:
## Description

Adds `swerve_controller_test`, an experimental swerve controller that smooths a
different set of quantities, plus the joystick modes that drive it.

`swerve_controller` slew-limits vx, vy and wz independently. The steering angles
are built from the *ratios* of those three, so whenever they ramp at different
rates the ratios drift and the wheels twitch — most visibly when only the
throttle moves and the turn geometry was meant to stay put.

This controller splits the twist into **shape** (which sets the steering angles,
and nothing else) and **magnitude** (which sets the wheel speeds, and nothing
else) and smooths those separately. A throttle change then touches magnitude
alone and the steering angles do not move at all.

The shape is a unit vector on a sphere whose pole is spin-in-place, written as
`theta` (crab direction) and `phi = atan(d * 1/R)`. Compactifying curvature
through `atan` is what makes spin-in-place reachable: `1/R` runs to infinity
there, but `phi` just reaches 90°, an ordinary point needing no special case.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [x] Configuration change

## Changes Made

- **`twist_shape.hpp`** — the shape/magnitude split: `decompose`/`recompose`, antipode resolution, and `ShapeSmoother` which rate-limits `theta`/`phi` in the sphere's own metric (the `cos(phi)` factor on `theta` frees it at the pole, otherwise exiting a spin gimbal-locks).
- **`swerve_controller_test`** — new ros2_control plugin. Reuses `SwerveKinematics` and `SlewRateLimiter` unchanged; `swerve_controller` is not touched.
- **Spin-in-place** — `Twist(0, 0, wz)` lands on the pole with no special case, and the transition into it is a continuous sweep that walks the turn centre in from infinity.
- **Angle-only commands** — a near-zero Twist steers the wheels with the drives held at zero (`park_speed`). Normalising recovers the same shape at any scale, so `1e-6` carries a usable direction.
- **Return-to-zero** — wheels home to straight after `idle_home_delay` of a genuinely empty command *and* a stopped rover. A near-zero Twist is not "empty", so parking the wheels at an angle still works.
- **Sticky branch selection** — a wheel at angle `a` with speed `+s` is identical to `a±180` with `−s`; the IK re-picked between them every cycle against a moving reference, which could flip mid-manoeuvre and send one wheel 213° round with its drive reversing. The choice is now continuous while driving, and re-picked freely while stopped (a 180° jump is free when no torque is being delivered, and buys the short path).
- **No `wz` clamp** — clamping yaw rate silently tightens the turn radius as speed drops. Magnitude and `phi` are clamped instead; turn radius is unbounded.
- **Joystick** — L2/R2 spin-in-place (`L2 − R2`, so a shared rest offset cancels); `controller_name` parameter; curvature mode now uses a *signed* speed so reversing keeps the turn centre on the same side, like a car; `max_curvature` replaces a doubled `scale_angular`.
- **Config** — `swerve_controller_test` blocks in both `controllers.yaml`. Sim runs it as the active controller (**temporary**); hardware loads it inactive alongside `swerve_controller`, selected via `controller_name`.

## Testing

- [ ] All existing tests pass


OShevchuk-UCU (changes requested):
The shape/magnitude decomposition looks like a promising approach, and the underlying `(m, theta, phi)` representation appears mathematically consistent.

Before merging, I think a few areas would benefit from additional attention, particularly because this controller may be used on the physical rover:

* The `/cmd_vel` callback and the controller `update()` loop currently access the same command fields from different execution contexts. It would be safer to pass the complete command through a real-time-safe buffer so that `update()` always receives a consistent Twist snapshot.
* The compact-mode service similarly modifies the kinematics state while `update()` may be using it. Applying this change through the real-time loop would avoid a possible race.
* The current `stopped` condition is based on the internal magnitude limiter rather than measured rover or wheel velocity. This could allow `snap()` or immediate wheel homing while the rover is still physically moving, especially with `idle_home_delay: 0.0`.
* The alignment scale is calculated from measured steering angles, while the drive direction correction uses the integrated commanded angles. Using the same angle source for both would make the behavior easier to reason about and reduce the risk of incorrect drive commands when steering feedback lags.
* Non-finite Twist values should be rejected before they reach the kinematics or hardware command interfaces.
* Since this controller is described as experimental, it may be safer to keep it opt-in rather than making it the default controller in simulation and teleoperation until the main transition cases have been covered by tests.

It would also be helpful to add tests covering:

* concurrent `/cmd_vel` updates and controller execution;
* compact-mode changes while the controller is active;
* stopping and coasting without early `snap()` or homing;
* pure spin and transitions into and out of spin;
* forward-to-reverse transitions;
* throttle-only changes leaving steering angles unchanged;
* steering feedback lag and invalid feedback;
* NaN/Inf inputs and invalid parameters;
* joystick trigger direction;
* active-controller selection during launch.

The mathematical idea looks worth continuing with, but I would prefer to address the real-time handoff and standstill detection before using this controller as the default path.

OShevchuk-UCU (changes requested):
## Review

Overall, the new shape/magnitude approach looks well thought out. Separating steering geometry from speed addresses the steering twitch caused by independently smoothing `vx`, `vy`, and `wz`, and the implementation has good coverage around spin-in-place, reverse, standstill detection, feedback failures, and compact mode.

I have one question about the intended controller architecture before considering the controller selection itself an issue:

**Is the goal to eventually replace `swerve_controller` with `swerve_controller_test` as the main controller?**

The code and documentation currently describe `swerve_controller_test` as experimental and opt-in, while the hardware bringup effectively moves toward using the new controller instead of keeping both available. If replacing the old controller is intentional, I think the configuration makes sense, but the documentation should reflect that direction.

The concrete issue I see is the joystick/controller selection. The shared `joy.yaml` points to `swerve_controller_test`, while another swerve controller may already be active. The joystick can then try to activate the new controller without deactivating the currently active one, or send compact-mode commands to a controller that is not active. Controller selection and switching should be made consistent so only the intended swerve controller owns the interfaces at a time.

There are also several non-blocking points worth considering:

* `ShapeSmoother` applies separate θ/φ limits rather than true geodesic interpolation, so some comments may overstate the mathematical guarantee.
* The `max_phi_rate` derivation does not fully bound steering demand when θ and φ change simultaneously.
* `resolve_antipode()` has no hysteresis around `shape_dot ≈ 0`.
* `dt` has no upper bound, so a delayed control cycle could produce a large one-cycle step.

The core controller design looks promising; the main thing I would resolve before merge is making the controller-selection behavior unambiguous and consistent across hardware, simulation, and joystick control.

yuriifito:
Now `swerve_controller_test` is the main controller for wheels. I keep `swerve_controller` just for a case of regression

PR #516, opened 2026-08-06
Feature/arm/replace meshes

milka-side:
## Description
Updates the arm to the new (v2) mesh set and fixes everything that broke or needed adjusting because of it — the gripper's rest-pose orientation, the wrist camera's mount position/orientation, several collisions that were silently blocking all Servo motion, and gamepad/keyboard teleop's roll behavior.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update
- [ ] Configuration change

## Changes Made
- Replaced all arm meshes with the new v2 set and updated joint origins/axes in `arm_macro.xacro` to match.
- Rotated the gripper 180° at its rest pose (`arm_wrist_2_end_effector_joint` origin) and re-mounted the wrist camera on top of the gripper, facing forward with the correct up axis, compensated for the gripper's rotation.
- Fixed the simulated camera actually seeing its own placeholder body: the Gazebo `<sensor>` renders from `arm_camera_link`'s own origin (not the separate `arm_camera_optical_frame`), so its `<pose>` needed its own forward offset, along the correct axis.
- Added an `arm_jaw_gripper_finger_right_joint` `ros2_control` interface — without it, the planning scene monitor considered the robot state permanently incomplete and silently blocked all Servo/MoveIt motion.
- Added SRDF collision exceptions for gripper-finger self-collision (both fingers vs `end_effector`, finger vs finger at their closed rest position, and camera vs both fingers) — these were constantly triggering `DECELERATE_FOR_COLLISION`, which looked like a permanent kinematic singularity but wasn't.
- Fixed roll (L2/R2, `i`/`k`) jogging the whole arm instead of just the last joint: it's now sent as a joint-space `JointJog` directly to `arm_wrist_2_end_effector_joint` instead of a Cartesian `angular.x` twist, which needed the rest of the arm to compensate for TCP's offset from that joint's axis.
- Fixed the gripper mimic joint: kept `multiplier="-1.0"` on the left finger but gave it its own `[-finger_stroke, 0]` limit range (was sharing `[0, finger_stroke]` with the right finger, which made any positive right-finger position produce an out-of-range left value).
- Aligned `arm_camera_optical_joint`'s offset with the Gazebo `<sensor>`'s own `0.045m` pose offset, so the published TF and the actual rendered camera origin agree.
- Replaced `arm_tcp_joint`'s stale offset (`0 -0.23 0`, left over from the pre-v2 design where the last joint's axis was Y, not Z) with a value computed from the fingertip midpoint in `arm_jaw_gripper_finger_v2.stl`'s own mesh geometry.

## Testing
- [ ] All existing tests pass

### How to Test
1. `ros2 launch arm_moveit_config demo.launch.py use_rviz:=false` (or the real hardware/Gazebo bringup) and confirm no `DECELERATE_FOR_COLLISION` or "incomplete robot state" warnings appear at the default safe pose.
2. `ros2 run joy joy_node` + `ros2 run arm_tasks gamepad_servo_node` (or `keyboard_servo_node` for keyboard) — move to safe pose, then jog roll (L2/R2 or `i`/`k`) and confirm only the last wrist joint turns.
3. View `/camera/image_raw` and confirm the camera's own body is no longer visible in its own frame.

## Additional Notes
The gripper's 180° rest-pose rotation was a deliberate visual/design change, not a kinematic workaround — no joint axes were altered for it. The camera's exact mount position was chosen interactively against a reference screenshot; flag if it still needs adjusting once mounted on the real hardware.

This PR changes several joint origins, axes, and limits alongside the v2 mesh swap, so previously-tuned poses/calibration should be revalidated before relying on them.


OShevchuk-UCU (changes requested):
## Review summary

The PR addresses several real issues introduced by the v2 arm meshes, and the Servo roll change to `JointJog` looks well structured. However, there are a few geometry/kinematics issues that should be resolved before merge.

The main concern is the new gripper mimic configuration: the left finger uses `multiplier="-1.0"` while both finger joints have limits `[0, finger_stroke]`. This means any positive right-finger position produces a negative left-finger joint value outside its declared limits. Please either use a positive mimic multiplier with the current axes or adjust the left joint limits accordingly.

The camera transforms also need another check. The Gazebo sensor is now offset by `0.045 m` from `arm_camera_link`, while `arm_camera_optical_frame` remains at the link origin. This can make the published TF extrinsics disagree with the actual rendered camera position.

Similarly, because the end-effector orientation and several wrist transforms changed significantly, the unchanged `arm_tcp_joint` offset should be revalidated against the CAD/tool-tip position. A simple FK check at the zero pose would be useful here.

The SRDF collision exceptions should also be verified carefully, especially finger↔finger and camera↔finger. These may fix false `DECELERATE_FOR_COLLISION` states, but broad collision disables can also hide genuine collisions if the new collision meshes or camera placeholder geometry are inaccurate.

Finally, this PR changes multiple joint origins, axes, and limits, so I would not classify it as purely non-breaking. Please mark it as a breaking/configuration change and mention that existing poses/calibration may need to be revalidated.

### Verdict

**Request changes**

Before merge I would address:

* gripper mimic/limit consistency;
* Gazebo camera pose vs optical-frame TF;
* TCP verification after the new EE/wrist transforms;
* PR classification / migration note.

The remaining Servo UX, collision-matrix tightening, and full gripper controller integration can reasonably be handled as follow-ups.

milka-side:
SRDF collision exceptions — verified this rather than just asserting it. Computed the STL bounding boxes for both finger meshes at their closed (joint=0) rest position, and they overlap on all three axes — expected behavior for a two-finger gripper when closed, not a modeling defect. The camera placeholder is likewise mounted flush against the gripper body by design, with a small (~2.5mm) intentional overlap. All three new exceptions are narrow and pairwise (camera↔finger_left, camera↔finger_right, finger_left↔finger_right); collision checking against the rest of the arm, and against external objects in the planning scene, is unaffected and remains fully active.

OShevchuk-UCU (approved):
Thanks for addressing the previous comments. The blocking issues are resolved now, and the SRDF collision exceptions look reasonable with the provided explanation. Looks good to me.

